PR #32, opened 2016-10-18
adding support for multiple json files

sharifelgamal:


reconciletags/reconciletags.py
```diff
@@ -65,15 +65,16 @@ def main():
     parser = argparse.ArgumentParser()
     parser.add_argument('--dry-run', dest="dry_run",
                         action='store_true', default=False)
-    parser.add_argument('file',
+    parser.add_argument('files',
```
dlorenc: I think you want to use something like nargs='+' here to properly support a repeated argument:
https://docs.python.org/3/library/argparse.html#example

reconciletags/reconciletags.py
```diff
@@ -51,29 +51,33 @@ def reconcile_tags(self, data, dry_run):
         # reconciler.
         self.call('gcloud config list', False)
         for repo, images in data.items():
-            logging.debug(self.get_existing_tags(repo))
+            if not dry_run:
+                logging.debug(self.get_existing_tags(repo))
```
dlorenc: I think this is fine to run during dry-run mode. Any reason you dropped it?

sharifelgamal: Mostly to reduce logging output while testing.

PR #40, opened 2016-10-29
Allow for repositories to be part of multiple registries

sharifelgamal:
@dlorenc

### DIFF
--- a/reconciletags/reconciletags.py
+++ b/reconciletags/reconciletags.py
@@ -8,6 +8,7 @@ If there are no changes that api call is no-op.
 import argparse
 import json
 import logging
+import os
 import subprocess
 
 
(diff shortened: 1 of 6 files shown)


reconciletags/reconciletags.py
```diff
@@ -50,15 +50,17 @@ def reconcile_tags(self, data, dry_run):
         # want to see config regardless of whether we actually run the
         # reconciler.
         self.call('gcloud config list', False)
-        for repo, images in data.items():
-            logging.debug(self.get_existing_tags(repo))
+        for project in data['projects']:
+            for registry in project['registries']:
+                full_repo = registry + '/' + project['repository']
```
dlorenc: Could you use os.path.join here? It would make it a little safer in case someone includes a / at the end of registry by accident.

sharifelgamal: Done.

PR #45, opened 2016-11-03
fixing reconciler to handle multiple registries properly

sharifelgamal:
@dlorenc

### DIFF
--- a/reconciletags/reconciletags.py
+++ b/reconciletags/reconciletags.py
@@ -52,16 +52,17 @@ class TagReconciler:
         # reconciler.
         self.call('gcloud config list', False)
         for project in data['projects']:
-            default_registry = project['registries'][0]
-            for registry in project['registries']:
+            default_registry = project['base_registry']
+            registries = project['additional_registries']
+            registries.append(default_registry)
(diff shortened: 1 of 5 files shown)


reconciletags/reconciletags.py
```diff
@@ -52,12 +52,16 @@ def reconcile_tags(self, data, dry_run):
         # reconciler.
         self.call('gcloud config list', False)
         for project in data['projects']:
+            default_registry = project['registries'][0]
```
dlorenc: Would it make sense to break this out to it's own field?
base_regsitry: "gcr.io"

then additional_registries: [...]

reconciletags/reconciletags.py
```diff
@@ -52,12 +52,17 @@ def reconcile_tags(self, data, dry_run):
         # reconciler.
         self.call('gcloud config list', False)
         for project in data['projects']:
-            for registry in project['registries']:
+            default_registry = project['base_registry']
+            registries = project['additional_registries']
```
dlorenc: Maybe make this a ".get" check in case someone leaves it off. This field should be optional.

PR #55, opened 2016-11-09
adding tests for data consistency

sharifelgamal:
@dlorenc


reconciletags/data_integrity_test.py
```diff
@@ -0,0 +1,45 @@
+"""Tests for data_integrity."""
```
dlorenc: Could you elaborate a bit more here?

PR #71, opened 2016-12-05
Make the severity to trigger on configurable.

dlorenc:


dlorenc:
cc @sharifelgamal @tejal29

PR #78, opened 2016-12-20
adding permissions check to file existence tests

### LINKED ISSUE #77: Verify file permissions as part of fileExistenceChecks
It would be great to be able to assert file permissions as part of fileExistenceChecks. Without this, it doesn't do much good to verify that a file exists if it's not readable/writable/executable/etc by the default docker user.

sharifelgamal:
Fixes #77 

@dlorenc @nkubala

### DIFF
--- a/structure_tests/README.md
+++ b/structure_tests/README.md
@@ -5,9 +5,9 @@ This code builds an image which serves as a framework to run structure-based tes
 
 To use this test image with any cloudbuild, add the following build step to the **end** your container build config (cloudbuild.yaml or cloudbuild.json):
 
-	name: gcr.io/gcp-runtimes/structure_test
-	args:
-		- <your_target_image>
+              name: gcr.io/gcp-runtimes/structure_test
+              args:
+                  - <your_target_image>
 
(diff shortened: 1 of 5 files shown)


structure_tests/file_existence_test_v1.go
```diff
@@ -23,6 +23,7 @@ type FileExistenceTestv1 struct {
 	Path        string // file to check existence of
 	IsDirectory bool   // whether or not the path points to a directory
 	ShouldExist bool   // whether or not the file should exist
+	Permissions string // expected permission string of the file
```
dlorenc: Could you add an example of the format of the string?

sharifelgamal: Done.

structure_tests/README.md
```diff
@@ -75,7 +77,8 @@ Example:
 		"name": "Root",
```
dlorenc: This diff is showing up with tabs instead of spaces, could you keep it as spaces?

sharifelgamal: Fixed.

PR #87, opened 2017-01-18
removing finding latest digest and adding logging

### LINKED ISSUE #86: Update gcloud and simplify the security test
containers describe should now be able to resolve digests, so we don't need to do that ourselves.

We can also add in more logging now of the actual vulnerabilities.

sharifelgamal:
Fixes #86

### DIFF
--- a/security_check/main.py
+++ b/security_check/main.py
@@ -58,7 +58,7 @@ def _check_image(image, severity, whitelist):
             logging.info('Found %s unpatched vulnerabilities in %s. Run '
                          '[gcloud beta container images describe %s] '
                          'to see the full list.',
-                          len(unpatched), image, image)
+                         len(unpatched), image, image)
         for vuln in unpatched.values():
             logging.info(pprint.pformat(vuln))
     return unpatched
--- a/security_check/main.py
+++ b/security_check/main.py
@@ -31,15 +31,23 @@ def _run_gcloud(cmd):
     return json.loads(output)
 
 
+def _find_base_image(image):
+    parsed = _run_gcloud(['describe', image])
+    img = parsed.get('image_analysis')
+    if img:
+        base_img_url = img[0]['base_image_url']
+        return base_img_url[len('https://'):base_img_url.find('@')]
+
+
 def _check_for_vulnz(image, severity, whitelist):
-    global _BASE_IMAGE
     unpatched = _check_image(image, severity, whitelist)
     if not unpatched:
         return unpatched
 
+    base_image = _find_base_image(image)
     base_unpatched = {}
-    if _BASE_IMAGE:
-        base_unpatched = _check_image(_BASE_IMAGE, severity, whitelist, True)
+    if base_image:
+        base_unpatched = _check_image(base_image, severity, whitelist)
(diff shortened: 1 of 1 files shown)


security_check/main.py
```diff
@@ -44,31 +43,27 @@ def _check_image(image, severity, whitelist):
...
+            unpatched[vuln['vulnerability']] = vuln
 
     if unpatched:
-        base_unpatched = 0
+        base_unpatched = {}
         img = parsed.get('image_analysis')
         if img:
             base_img_url = img[0]['base_image_url']
             base_image = base_img_url[len('https://'):base_img_url.find('@')]
             base_unpatched = _check_image(base_image, severity, whitelist)
```
dlorenc: it looks like this might recursively check all base images up to the root. Is there anything to prevent that?

security_check/main.py
```diff
@@ -44,31 +43,27 @@ def _check_image(image, severity, whitelist):
...
-        unpatched -= base_unpatched
-        logging.info('Found %s unpatched vulnerabilities in %s. Run '
-                     '[gcloud beta container images describe %s] '
-                     'to see the full list.',
-                     unpatched, image, full_name)
+        for vuln in base_unpatched.keys():
+            del unpatched[vuln]
```
dlorenc: Can we log the vulns we're skipping because of the base? We might want to make that behavior a flag too.

sharifelgamal: Done. I'm gonna forego the flag since we're the only users of this script currently.

security_check/main.py
sharifelgamal: What would be desirable behavior here? Iterate up until root first then check?

sharifelgamal: refactored how we check the base image. PTAL.

security_check/main.py
```diff
@@ -30,12 +31,43 @@ def _run_gcloud(cmd):
...
+
+
+def _check_for_vulnz(image, severity, whitelist):
+    unpatched = _check_image(image, severity, whitelist)
+    if not unpatched:
+        return unpatched
+
+    base_image = _find_base_image(image)
+    base_unpatched = {}
+    if base_image:
```
dlorenc: Could you log if we can't find the base image?

security_check/main.py
```diff
@@ -30,12 +31,43 @@ def _run_gcloud(cmd):
...
+        base_unpatched = _check_image(base_image, severity, whitelist)
+    for vuln in base_unpatched.keys():
+        logging.info('Vulnerability %s exists in the base image. Skipping.',
+                     vuln)
+        del unpatched[vuln]
```
dlorenc: Could it be possible for the base image to contain a vuln, but not the child image? I guess if the child image explicity does an apt-get upgrade it could happen. I think in that case this line would fail.

Maybe it would be better to iterate through the unpatched list, and re-collect the vulns into a new object if they're not present in the base?

sharifelgamal: You're absolutely right, that can happen. Will fix.

security_check/main.py

PR #106, opened 2017-02-22
fixing reconciler bug

sharifelgamal:


sharifelgamal:
Agreed, that just requires a refactor.

PR #123, opened 2017-03-03
Use a different method of mounting the cloudbuild workspace

duggelz:


duggelz:
CC @nkubala @dlorenc 

This fixes the local_cloudbuild problem.  The important lines are 81-83 in run.sh.  Let me know if this seems reasonable to you.

For some reason I can't select reviewers, not sure if I don't have the right permission or if I just clicked the wrong button somewhere.

sharifelgamal:
LGTM

structure_tests/run.sh
```diff
@@ -74,4 +76,8 @@ if [ -z "$IMAGE_NAME" ]; then
 	usage
 fi
 
-docker run --privileged=true -v /workspace:/workspace --entrypoint="$ENTRYPOINT" "$IMAGE_NAME" -c "$CMD_STRING"
+# Get id of container we're currently in.  This method is
+# system-specific but probably better than looking at $HOSTNAME.
+THIS_CONTAINER=$(basename $(head -1 /proc/self/cgroup))
```
nkubala: this seems to work as far as I can tell. just curious why you think it's better than looking at hostname? also seems to work in all cases for me.

structure_tests/tests/structure_test_tests.sh
```diff
@@ -20,49 +20,45 @@
...
 FILE="debian_failure_test.json"
 envsubst < cloudbuild.yaml.in > cloudbuild.yaml
-gcloud beta container builds submit . --config=cloudbuild.yaml
-if [ "$?" -ne "1" ]
-then
-  echo "Failure case test failed"
-  failures=$((failures + 1))
-fi
+gcloud beta container builds submit . --config=cloudbuild.yaml && \
+  failure "Failure case test failed"
```
nkubala: will these expected failures get added to the total failure count at the end? if so that seems a little weird to me.

structure_tests/run.sh
duggelz: It's a bit far-fetched, but you could do `docker run --name=foo --hostname=bar` which would mess up just looking at `$HOSTNAME`.  Also the hostname gets truncated at 12 characters so [collisions](https://github.com/docker/docker/issues/28260) can happen.

`THIS_CONTAINER=$HOSTNAME` is a whole lot simpler though.

structure_tests/tests/structure_test_tests.sh
duggelz: I wondered if I was being too clever here.  And the answer seems to be yes.  The expected failure case has `&&` instead of `||` so that the count is properly zero at the end, but it's easy to miss that difference.  I've changed that back to the way it was.

structure_tests/run.sh
nkubala: ahh makes sense.

duggelz:
@nkubala You'll have to merge this for me, at your convenience.

meltsufin:
@nkubala Can you release `gcr.io/gcp-runtimes/structure_test` with this change?

nkubala:
@meltsufin ahh yep, thanks for the reminder.

duggelz:
We're in a production freeze until GCP Next is done, though.

meltsufin:
I thought the freeze only applied to runtime images themselves, not the supporting test infra.

nkubala:
I believe @meltsufin is right, the test-related images shouldn't be frozen since they don't affect end users.

PR #133, opened 2017-03-17
Add containerFromRun for exec instruction

huyhg:
The container name will be derived from the specified run
instruction. Before, the user had to specify a hard-coded name.


huyhg:
Ping.

PR #139, opened 2017-03-31
switch template builder to accept a file instead of a directory

sharifelgamal:


sharifelgamal:
That is indeed the plan.

PR #149, opened 2017-04-18
Add verify flag to dockerfiles

wgrzelak:
It may be useful to verify dockerfiles.

### DIFF
--- a/versioning/scripts/dockerfiles/main.go
+++ b/versioning/scripts/dockerfiles/main.go
@@ -7,6 +7,8 @@ import (
 	"bytes"
 	"flag"
 	"io/ioutil"
+	"log"
+	"os"
 	"path/filepath"
 	"text/template"
 
@@ -33,6 +35,8 @@ func check(e error) {
 
 func main() {
 	templateDirPtr := flag.String("template_dir", "templates", "Path to directory containing Dockerfile.template")
+	verifyPtr := flag.Bool("verify", false, "Verify dockerfiles")
+	flag.Parse()
 
 	var spec versions.Spec
 	spec = versions.LoadVersions("versions.yaml")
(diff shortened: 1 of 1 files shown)


versioning/scripts/dockerfiles/main.go
```diff
@@ -25,6 +27,40 @@ func writeDockerfile(version versions.Version, data []byte) {
...
+		dockerfile, err := ioutil.ReadFile(path)
+		check(err)
+
+		foundDockerfile[path] = true
+
+		if string(dockerfile) == string(data) {
+			log.Printf("%s: OK", path)
+		} else {
+			failureCount++
+			log.Printf("%s: FAIL", path)
```
huyhg: s/FAIL/FAILED/

versioning/scripts/dockerfiles/main.go
```diff
@@ -25,6 +27,40 @@ func writeDockerfile(version versions.Version, data []byte) {
...
+		}
+	}
+
+	err := filepath.Walk(".", func(path string, info os.FileInfo, err error) error {
```
huyhg: As it turns out, the PHP repo (on gerrit under limani-php) has `Dockerfile`s that are not included in `versions.yaml`. Let's remove this check for now. We can also extend `versions.yaml` to have an ignored list of `Dockerfile`s, but that doesn't have to be part of this PR.

versioning/scripts/dockerfiles/main.go
```diff
@@ -33,6 +69,8 @@ func check(e error) {
 
 func main() {
 	templateDirPtr := flag.String("template_dir", "templates", "Path to directory containing Dockerfile.template")
+	verifyPtr := flag.Bool("verify-only", false, "Verify dockerfiles")
```
huyhg: s/verify-only/verify_only/

PR #150, opened 2017-04-18
Prepare for bazel-ifying the entire repo.

dlorenc:
This moves WORKSPACE to the top level and renames some relative paths inside docgen.


dlorenc:
> Awesome! Please do the same to versioning

Ah nice, done.

PR #155, opened 2017-04-21
Fix single file mounting for kubernetes

huyhg:
configmap should be mounted on a directory. Keys in the configmap
will become file names in that directory.


huyhg:
@dlorenc could you help review this quick change while David is out? Thanks!

PR #162, opened 2017-05-01
fixing security tests to account for different bases

sharifelgamal:


security_check/main.py
```diff
@@ -51,7 +51,10 @@ def _find_base_image(image):
     img = parsed.get('image_analysis')
     if img:
         base_img_url = img[0]['base_image_url']
-        return base_img_url[len('https://'):base_img_url.find('@')]
+        base_img = base_img_url[len('https://'):base_img_url.find('@')]
```
dlorenc: Could you make this a tiny bit more generic, and maybe have a map of substitutions?

sharifelgamal: done

security_check/main.py
```diff
@@ -39,6 +39,19 @@
...
+    'l.gcr.io/google': 'gcr.io/google-appengine'
+}
+
+
+def _sub_image(image):
+    for k in _SUB_MAP.keys():
+        if k in image:
+            image = image.replace(k, _SUB_MAP.get(k))
+
+    return image
```
dlorenc: Ah, this is more annoying than I thought it would be since we don't parse the image into a registry and a repo :(

Do you think that would be easier? Basically split on / and the first two parts are the registry/repo then you can combine them, and do a simple lookup?

sharifelgamal: That makes sense.

security_check/main.py
```diff
@@ -39,6 +39,17 @@
...
 }
 
+_SUB_MAP = {
+    'launcher.gcr.io/google': 'gcr.io/google-appengine',
+    'l.gcr.io/google': 'gcr.io/google-appengine'
+}
+
+
+def _sub_image(image):
+    repoAndImage = image.rsplit('/', 1)
```
dlorenc: This might be a little cleaner as:

repo, image = image.rsplit('/', 1)

sharifelgamal: I forgot Python did that. good call.

security_check/main.py
```diff
@@ -39,6 +39,17 @@
...
 }
 
+_SUB_MAP = {
+    'launcher.gcr.io/google': 'gcr.io/google-appengine',
+    'l.gcr.io/google': 'gcr.io/google-appengine'
+}
+
+
+def _sub_image(image):
+    repo, image = image.rsplit('/', 1)
```
dlorenc: Could you add some logging to this? It might be confusing to users at first.

sharifelgamal: Let me know if this sort of logging was what you had in mind.

PR #167, opened 2017-05-03
Support excluding specific tests per version

huyhg:


versioning/scripts/cloudbuild/main.go
```diff
@@ -188,13 +190,40 @@ func readTests(testsDir string) (tests []string) {
...
 }
 
+func filterTests(structureTests []string, functionalTests []string, version versions.Version) (outStructureTests []string, outFunctionalTests []string) {
+	included := make(map[string]bool, len(structureTests)+len(functionalTests))
+	for _, test := range append(structureTests, functionalTests...) {
+		included[test] = true
+	}
+	for _, excluded := range version.ExcludeTests {
+		if !included[workspacePrefix+excluded] {
+			log.Fatalf("No such test to exclude: %s", excluded)
```
jvolkman: Why fatal rather than a warning?

huyhg: Because nobody will see the warning message ;) I want this to abort the build because this is likely a typo. It doesn't make sense to have an excluded test that doesn't exist.

PR #170, opened 2017-05-09
Fix typo in versioning/README.md

wgrzelak:


huyhg:
These are not really typos--we want to codify only the `Dockerfile` part, leaving the "s" in normal font,

PR #180, opened 2017-05-17
Add a bazel rule definition for structure tests.

dlorenc:
cc @r2d4 @mattmoor


mattmoor:
Maybe this is tied to this repo more broadly, but I think this would make a nice addition to `bazelbuild/rules_docker`.  I'd love to move some of the stuff currently in `build_test.sh` out of bash and into this...  No rush, but something we should discuss.

dlorenc:
The nice thing about build_test.sh is that it works off the tars directly and doesn't need a docker daemon.

mattmoor:
i did not realize it had that dep, does it need it?

dlorenc:
Not necessarily for the file existence/content checks, but it also supports running commands and checking the stdout/stderr/return codes. Those would require a docker daemon.

PR #212, opened 2017-06-27
Allow structure tests to be run in parallel.

dlorenc:
This should fix https://github.com/GoogleCloudPlatform/distroless/issues/65


dlorenc:
cc @nkubala

PR #218, opened 2017-06-29
Fix import path.

dlorenc:


dlorenc:
cc @abbytiz @cftorres @nkubala

dlorenc:
This is passing now :)

PR #219, opened 2017-06-30
refactoring the security check

sharifelgamal:


security_check/__init__.py
```diff
@@ -0,0 +1,6 @@
+def main(argv):
```
dlorenc: Can this file just be empty?

dlorenc: Well, with a copyright header :)

security_check/main.py
```diff
@@ -158,9 +158,19 @@ def _check_vuln_is_valid(vuln):
                  fixed_version)
     return False
 
+def _security_check(image, severity=_MEDIUM, whitelist_file='whitelist.json'):
+    apt_pkg.init()
+    logging.basicConfig(level=logging.DEBUG)
```
dlorenc: Move the logging config call back up into main.

security_check/__init__.py
sharifelgamal: Done.

security_check/main.py

security_check/main.py
```diff
@@ -159,8 +159,19 @@ def _check_vuln_is_valid(vuln):
     return False
 
 
-def _main():
+def _security_check(image, severity=_MEDIUM, whitelist_file='whitelist.json'):
```
dlorenc: One last nit, you can drop the leading underscore since we're exposing this more broadly now.

PR #221, opened 2017-06-30
Pip Differ

cftorres:


iDiff/differs/pipDiff.go
```diff
@@ -0,0 +1,117 @@
...
+	"path/filepath"
+	"regexp"
+	"strconv"
+
+	"github.com/GoogleCloudPlatform/runtimes-common/iDiff/utils"
+	"github.com/golang/glog"
+)
+
+// layers of two different images.
+func PipDiff(d1file, d2file string) (string, error) {
```
abbytiz: There is significant overlap in the differs in terms of having a driver like this which calls a differ specific getPackages function and then gets the diff.  Maybe in a future iteration/refactor we will want to move this repeated code so that it's no longer repeated.

iDiff/differs/pipDiff.go
```diff
@@ -0,0 +1,117 @@
...
+	d2, err := utils.GetDirectory(d2file)
+	if err != nil {
+		glog.Errorf("Error reading directory structure from file %s: %s\n", d2file, err)
+		return "", err
+	}
+
+	dirPath1 := d1.Root
+	dirPath2 := d2.Root
+	pack1 := getPythonPackages(dirPath1)
+
```
abbytiz: nit: extra newline

iDiff/differs/pipDiff.go
cftorres: yes, was thinking about this too and have it on my TODO list

iDiff/differs/pipDiff.go
```diff
@@ -0,0 +1,117 @@
...
+		if match != "" {
+			return match, true
+		}
+	}
+	return "", false
+}
+
+// TODO: Eventually, this would make use of the shallow JSON and be diffed
+// with that of another image to get only the layers that have changed.
+func getImageLayers(pathToImage string) []string {
```
abbytiz: I think you could similarly do this with the BuildLayerTarget function by passing an empty string for the target.

aaron-prindle:
LGTM.
In another PR add integration tests for this differ once the URL and travis script PRs go in.  @abbytiz you should also add a PR for the apt-differ once the testing PRs go in.

PR #225, opened 2017-06-30
Node differ

abbytiz:
Handles npm installed node module diffing


iDiff/utils/package_diff_utils.go
```diff
@@ -1,11 +1,31 @@
...
+// MultiVersionPackageDiff stores the difference information between two images which could have multi-version packages.
+type MultiVersionPackageDiff struct {
+	Image1    string
+	Packages1 map[string]map[string]PackageInfo
+	Image2    string
+	Packages2 map[string]map[string]PackageInfo
+	InfoDiff  []MultiVersionInfo
+}
+
+// MultiVersionInfo stores the information for one multi-version package in two different images.
```
cftorres: just to clarify-- so is it true that multi-version packages would exist across images, not within? because in your tests, it looks like there's a multi-version package example within one image...

abbytiz: The multi-version additions are intended to handle the case of multiple versions of the same package within one image, which then need to be compared with another image's packages which may also have multiple versions.

iDiff/utils/format_utils.go
```diff
@@ -9,3 +15,43 @@ func JSONify(diff interface{}) (string, error) {
...
+	}
+
+	if err := masterTmpl.Execute(os.Stdout, diff); err != nil {
+		log.Fatal(err)
+	}
+	return nil
+}
+
+func OutputMulti(diff MultiVersionPackageDiff) error {
+	const master = `Packages found only in {{.Image1}}:{{range $name, $value := .Packages1}}{{"\n"}}{{print "-"}}{{$name}}{{end}}{{"\n"}}
```
aaron-prindle: It might also be useful in the future if users can pass in a custom Go template for their output.  Essentially allow this string as an input.  Could make testing easier as well.

aaron-prindle:
LGTM

PR #226, opened 2017-07-05
Fixing apt size parsing and adding tests.

abbytiz:
Noticed size wasn't being populated, this fixes that issue and adds tests for those cases.


aaron-prindle:
LGTM.  With all the package information in structs, you could try to output the structs in package_diff_utils.go as serialized values and that should be a good start to outputting json/go-templated information.

PR #228, opened 2017-07-05
Json output

abbytiz:
Hist and dir don't currently return JSON's.  I have side-stepped the output formatting so that the apt differ now returns the string version of the JSON for the PackageDiff struct.

### DIFF
--- a/iDiff/differs/differs.go
+++ b/iDiff/differs/differs.go
@@ -1,7 +1,10 @@
 package differs
 
 import (
+	"bufio"
+	"bytes"
 	"errors"
+	"io"
 	"os"
 
 	"github.com/GoogleCloudPlatform/runtimes-common/iDiff/utils"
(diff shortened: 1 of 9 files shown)


iDiff/differs/differs.go
```diff
@@ -13,20 +16,17 @@ var diffs = map[string]func(string, string) (string, error){
 	"apt":  AptDiff,
 }
 
-func Diff(arg1, arg2, differ string) (string, error) {
+func Diff(arg1, arg2, differ, output string) (string, error) {
 	if f, exists := diffs[differ]; exists {
 		if differ == "hist" {
 			return f(arg1, arg2)
```
cftorres: does the history differ not also have the option of outputting to a JSON file?

abbytiz: Good point - hist does not output to file at the moment.  I'm going to rework hist and dir to return JSONs when the flag is set, and similarly rework the apt differ so with the flag JSONs are returned, otherwise human-readable output is returned.  If the user wants to save the JSON to a file I think we should allow them to do that with the command line on their own.

aaron-prindle:
LGTM

PR #231, opened 2017-07-07
Added bash script for iDiff tests

cftorres:


iDiff/.iDiffTests.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+go run iDiff/main.go iDiff 0cb40641836c e7d168d7db45 dir -j > iDiff/tests/runs/busybox_diff_actual.json
```
abbytiz: To clarify, once the URL PR is in this will be changed to image URL's correct?

cftorres: yes!

aaron-prindle:
It seems this requires the URL PR to go in.  Get that merged first, then rewrite it using URLs pointing to public images.  Then the test will work like normal with Travis and we can use that to validate correctness.

aaron-prindle:
Once this is resubmitted and the tests past we can merge.

PR #232, opened 2017-07-10
Added image pull from URL functionality

cftorres:

### DIFF
--- a/iDiff/cmd/iDiff.go
+++ b/iDiff/cmd/iDiff.go
@@ -1,10 +1,8 @@
 package cmd
 
 import (
-	"bytes"
 	"errors"
 	"fmt"
-	"regexp"
 
 	"github.com/GoogleCloudPlatform/runtimes-common/iDiff/differs"
 	"github.com/golang/glog"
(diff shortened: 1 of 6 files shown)


iDiff/cmd/iDiff.go
```diff
@@ -34,67 +32,22 @@ func validateArgs(args []string) (bool, error) {
...
-	}
-	if !checkImageID(args[1]) {
-		valid = false
-		errMessage := fmt.Sprintf("Argument %s is not an image ID\n", args[1])
-		buffer.WriteString(errMessage)
-	}
-	if checkImageID(args[2]) {
-		valid = false
-		buffer.WriteString("Do not provide more than two image IDs\n")
-	} else if !checkDiffer(args[2]) {
```
abbytiz: I think we should try to keep some of the argument checking, especially the checkDiffer.

iDiff/utils/image_utils.go
```diff
@@ -72,3 +168,18 @@ func copyToFile(outfile string, r io.Reader) error {
 
 	return nil
 }
+
+func checkImageID(image string) bool {
```
abbytiz: This function may be more useful farther up the stack to help validate arguments.  Maybe in differs.go as we also expand to tar arguments. Also I think we had talked about trying to isolate Docker dependent things in this file.

iDiff/utils/image_utils.go
```diff
@@ -72,3 +168,18 @@ func copyToFile(outfile string, r io.Reader) error {
...
+	pattern := regexp.MustCompile("[a-z|0-9]{12}")
+	if exp := pattern.FindString(image); exp != image {
+		return false
+	}
+	return true
+}
+
+func checkImageTar(image string) bool {
```
abbytiz: This also seems useful farther up the stack.

cftorres: where else would you suggest?

abbytiz: Looking again it might make the most sense to just make these exported functions so they can be called elsewhere through the utils package. I can definitely imagine both of these functions being used in differs.go to handle function calling based on argument types or in argument verification of the command.

iDiff/utils/image_utils.go
```diff
@@ -2,47 +2,132 @@ package utils
...
+			return imageDigest, nil
+		}
 	}
-	err = ExtractTar(tarPath)
+	err := fmt.Errorf("Could not pull image %s", image)
+	return "", err
+}
+
+func pullImageFromRepo(cli client.APIClient, image string) (string, string, error) {
+	response, err := cli.ImagePull(context.Background(), image, types.ImagePullOptions{})
```
aaron-prindle: It is important to note that this may not work properly for all users.  This will invoke docker client commands(from the lib) to a docker daemon(pre-installed on the user's host machine) and the client/daemon might not have the correct version compatibility.  I believe Abby is working on a PR to resolve this but just going to note it here.

iDiff/cmd/iDiff.go
```diff
@@ -34,67 +32,22 @@ func validateArgs(args []string) (bool, error) {
...
 	} else if len(args) > 3 {
-		errMessage = "Too many arguments. Should have three: [IMAGE ID] [IMAGE ID] [DIFFER]."
+		errMessage = "Too many arguments. Should have three: [IMAGE] [IMAGE] [DIFFER]."
 		return false, errors.New(errMessage)
 	} else {
 		return true, nil
 	}
 }
 
-func checkArgType(args []string) (bool, error) {
```
aaron-prindle: I see that the tar/dir concept has been abstracted from each other to allow for tars to be used as input.  Adding path validation for the tar input might be useful for the checks as well

aaron-prindle:
LGTM.  Merge this when you're ready.

PR #236, opened 2017-07-13
Driver cleanup

abbytiz:


iDiff/cmd/root.go
```diff
@@ -1,22 +1,101 @@
...
+		if diff, err := differs.Diff(args[2], args[3], args[1], json); err == nil {
+			fmt.Println(diff)
+		} else {
+			glog.Error(err.Error())
+		}
 	},
 }
 
+func validateArgs(args []string) (bool, error) {
+	if validArgNum, err := checkArgNum(args); !validArgNum {
```
aaron-prindle: Check the error here as well.  It doesn't matter currently but the implementation for validArgNum might change.

iDiff/cmd/root.go
```diff
@@ -1,22 +1,101 @@
...
+		return false, err
+	}
+	if validArgType, err := checkArgType(args); !validArgType {
```
aaron-prindle: Check the error here as well. It doesn't matter currently but the implementation for validArgType might change.

aaron-prindle:
Also looks like the travis tests needs to be updated to not have 'iDiff' in the command

PR #238, opened 2017-07-14
Docker version errors

abbytiz:
Catches lack of docker on machine error and incompatible version for docker engine api


iDiff/utils/docker_utils.go
```diff
@@ -0,0 +1,172 @@
+package utils
```
cftorres: If possible to refactor code to move some of the stuff in image_utils that centers around Docker into this file, I think that would be better organization...

iDiff/differs/history.go
```diff
@@ -18,21 +19,34 @@ func History(img1, img2 string, json bool) (string, error) {
...
+		cli, err := client.NewEnvClient()
+		if err != nil {
+			return []string{}, err
+		}
+		history, err = cli.ImageHistory(ctx, img)
+		if err != nil {
+			return []string{}, err
+		}
+	} else {
+		glog.Info("Docker verison incompatible with api, shelling out to local Docker client.")
```
aaron-prindle: nit: version

iDiff/utils/docker_utils.go
```diff
@@ -0,0 +1,172 @@
...
+	var err error
+	var history []img.HistoryResponseItem
+	if !CheckImageID(image) {
+		imageID, _, err = pullImageCmd(image)
+		if err != nil {
+			return history, err
+		}
+	}
+	histArgs := []string{"history", "--no-trunc", imageID}
+	dockerHistCmd := exec.Command("docker", histArgs...)
```
cftorres: @nkubala @aaron-prindle should we consider using gcloud instead of docker commands for people who don't have Docker installed? Or is that circular kind of logic where people might not have gcloud...?

iDiff/utils/docker_utils.go
```diff
@@ -0,0 +1,172 @@
...
+			if err == io.EOF {
+				break
+			}
+			return "", "", err
+		}
+		event.Status = string(text)
+		events = append(events, event)
+	}
+
+	imageDigest, err := getImagePullResponse(image, events)
```
cftorres: Seems like from here down is repetitive code from image_utils? If there's a way to refactor to extract this out into a helper method that can be reused, that'd probably be better...

iDiff/utils/docker_utils.go
```diff
@@ -0,0 +1,172 @@
...
+		imageID, _, err = pullImageCmd(image)
+		if err != nil {
+			return history, err
+		}
+	}
+	histArgs := []string{"history", "--no-trunc", imageID}
+	dockerHistCmd := exec.Command("docker", histArgs...)
+	var response bytes.Buffer
+	dockerHistCmd.Stdout = &response
+	if err := dockerHistCmd.Run(); err != nil {
```
aaron-prindle: @r2d4 had mentioned that you might also be able to pull images using without using docker using something like https://github.com/google/containerregistry .  This could remove the docker dependency from the tool for CI.

aaron-prindle: Also https://github.com/projectatomic/skopeo

iDiff/utils/image_utils.go
```diff
@@ -43,9 +44,21 @@ func saveImageToTar(image string) (string, error) {
...
 		// If not an image tar already existing in the filesystem, create client to obtain image
-		imageTar, err := saveImageToTar(img)
+		// check client compatibility with Docker API
+		valid, err := ValidDockerVersion()
+		if err != nil {
+			return "", "", err
+		}
+		var imageTar string
+		if !valid {
+			glog.Info("Docker verison incompatible with api, shelling out to local Docker client.")
```
aaron-prindle: nit: version

iDiff/utils/docker_utils.go
r2d4: Theres not a lot of documentation on the containerregistry tool, but you can grab it at:
https://storage.googleapis.com/containerregistry-releases/v0.0.11/puller.par

Usage:
`$ puller.par --name ubuntu --tarball ubuntu.tar`
`$ docker load ubuntu.tar`

iDiff/utils/docker_utils.go
aaron-prindle: @cftorres gcloud seems to just pass the args to docker CLI https://cloud.google.com/sdk/gcloud/reference/docker.  I believe that all gcloud does is provide appropriate auth to the docker onboard.

aaron-prindle:
We can leave testing multiple docker versions to another PR.  We should also discuss potentially removing whatever docker dependencies we can.

iDiff/utils/docker_utils.go
```diff
@@ -0,0 +1,172 @@
...
+)
+
+// ValidDockerVersion determines if there is a Docker client of the necessary version locally installed.
+func ValidDockerVersion() (bool, error) {
+	cli, err := client.NewEnvClient()
+	if err != nil {
+		return false, fmt.Errorf("Docker client error: %s", err)
+	}
+	version := cli.ClientVersion()
+	if version == "1.31" {
```
aaron-prindle: In attempting to use this branch I am getting: `Error response from daemon: client is newer than server (client 
API version: 1.31, server API version: 1.24)`.  The error is related to client and server incompatibility.  This check doesn't seem sufficient, is cli.ClientVersion output the version of the lib:
https://github.com/moby/moby/blob/master/client/client.go#L236
so I think this will always be true.  I think you would need to get the server version and make sure 1.31 is compatible with that as the check.  Perhaps you can add a flag to use onboard docker and output to use the flag in case of an error or make that the default for now?

iDiff/utils/image_utils.go
```diff
@@ -70,6 +41,33 @@ func ImageToDir(img string, eng bool) (string, string, error) {
...
+
+func CheckImageURL(image string) bool {
+	pattern := regexp.MustCompile("^.+/.+(:.+){0,1}$")
+	if exp := pattern.FindString(image); exp != image || CheckTar(image) {
+		return false
+	}
+	return true
+}
+
 type Event struct {
```
cftorres: does this also make more sense in the docker_utils file since it is really only involved in Docker image pulls?

abbytiz: Definitely, good catch!

aaron-prindle:
LGTM

iDiff/differs/history.go
```diff
@@ -8,31 +8,45 @@ import (
...
 	if err != nil {
 		return []string{}, err
 	}
-	history, err := cli.ImageHistory(ctx, image)
-	if err != nil {
-		return []string{}, err
+	var history []image.HistoryResponseItem
+	if validDocker {
+		ctx := context.Background()
+		cli, err := client.NewEnvClient()
```
r2d4: The real issue here is that you end up setting up a client with a much higher client API version than your server if you don't have `DOCKER_API_VERSION` set.
Basically I think you're trying to avoid the code in `client.NewEnvClient()`

```golang
version := os.Getenv("DOCKER_API_VERSION")
if version == "" {
  version = api.DefaultVersion //1.31
}
```

I think you want this convenience method to set the client version appropriately.
```golang
// NegotiateAPIVersion updates the version string associated with this
// instance of the Client to match the latest version the server supports
func (cli *Client) NegotiateAPIVersion(ctx context.Context) {
```

https://github.com/moby/moby/blob/ff4f700f74450018f36d014f3cde0ff1b9c17fb3/client/client.go#L244-L249

iDiff/differs/history.go
```diff
@@ -8,31 +8,45 @@ import (
...
+		if err != nil {
+			return []string{}, err
+		}
+		history, err = cli.ImageHistory(ctx, img)
+		if err != nil {
+			return []string{}, err
+		}
+	} else {
+		glog.Info("Docker version incompatible with api, shelling out to local Docker client.")
+		history, err = utils.GetImageHistory(img)
```
r2d4: If you set the client API version appropriately for the server, you won't have to shell out

iDiff/cmd/iDiff.go
```diff
@@ -98,4 +99,5 @@ func checkArgType(args []string) (bool, error) {
 func init() {
 	RootCmd.AddCommand(iDiffCmd)
 	iDiffCmd.Flags().BoolVarP(&json, "json", "j", false, "JSON Output defines if the diff should be returned in a human readable format (false) or a JSON (true).")
+	iDiffCmd.Flags().BoolVarP(&eng, "eng", "e", false, "By default the docker calls are shelled out locally, set this flag to use the Docker Engine Client (version compatibility required).")
```
r2d4: If you set the client API version properly, you shouldn't need to shell out.  Then, you can use the client library everywhere.

iDiff/differs/differs.go
```diff
@@ -8,41 +8,41 @@ import (
...
-var diffs = map[string]func(string, string, bool) (string, error){
+var diffs = map[string]func(string, string, bool, bool) (string, error){
 	"hist": History,
 	"dir":  Package,
 	"apt":  AptDiff,
 	"node": NodeDiff,
 	"pip":  PipDiff,
 }
 
-func Diff(arg1, arg2, differ string, json bool) (string, error) {
```
r2d4: You could also extract the json encoding logic from the differs.  It gets duplicated for every differ, and it should be independent from the "diffing".  If a particular differ needs to encode json a special way, you might consider returning a more structured `ImageDiffOutput` struct that can json encode itself.

PR #271, opened 2017-07-26
Removing docker dependency from history calls

abbytiz:
Fixes errors produced when calling iDiff on tars and removes some docker dependencies.  #244 


abbytiz:
Just realized this doesn't quite work fully yet because some images have multiple .json files with the history.  Code fix to follow.

abbytiz:
Modified to be more specific in json file reading and to report an error if multiple files are found that are relevant.

PR #277, opened 2017-07-31
History Diff Integration Test

abbytiz:
@cftorres @nkubala @aaron-prindle


abbytiz:
I added a test that calls many differs (pip, apt, and node) with the history differ.  Package diff still works and nothing is skipped.  Maybe the docker dependency that was removed was unclear, but iDiff is still dependent on docker to pull the images and save that as .tar files.  It is no longer using docker to get the histories of the images but is getting those histories from the unpacked tar files.  The larger issue fixed was making sure that the history differ could be run on .tar files (which previously it could not).

PR #283, opened 2017-08-02
Fixed tabs in output template

cftorres:


aaron-prindle:
Does this maintain a fixed width for different sized entries (do the column positions zigzag due to size of a word)?  It might be useful in another PR to try to make the output columnar so that it is more readable for users, it's not clear to me that this PR would do that.

cftorres:
yes, previously it was zigzagging for some reason, but this seemed to fix that (if you recall from our demo just now, all the output was in fixed columns)

PR #310, opened 2017-08-17
Docgen: Fix docker compose generation

jprzychodzen:
* Additional backslash after image name if command is provided (fix)


docgen/lib/render/templates/README.md.tmpl
```diff
@@ -88,7 +88,7 @@ services:
   {{.Name}}:
     container_name: {{ .ContainerName }}
     image: {{.Image}}
-    {{- if gt (len .Arguments) 0}} \
+    {{- if gt (len .Arguments) 0}}
```
huyhg: Not sure how that had happened...

PR #322, opened 2017-08-28
Use the whole token for the filter

tmatsuo:
Rationale:

We used to use the `test_util.LOGGING_PREFIX` for the filter. It works most of the times, but if we run bunch of integration tests at the same time, the tests may start to fail because the log entry in question might not fit in a single page. We can just safely use the whole token for the filter.


tmatsuo:
@nkubala I'll appreciate if you can take a look.

tmatsuo:
Thanks! Can you merge this? I have no power to do it.

PR #327, opened 2017-08-31
Remove gcloud dependency from retagger

priyawadhwa:
This PR updates the WORKSPACE file and creates a BUILD file so that we can use bazel to pull in the containerregistry library, which is used to remove the google3 dependency in retagger (which is 
reconciletags.py)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,5 +35,4 @@ script:
  - ./.gofmt.sh
  - ./.buildifier.sh
  - bazel test reconciletags:reconciletags_test
- - bazel test reconciletags:reconciletags_e2e_test
  - bazel clean && bazel build //...
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,5 +34,5 @@ script:
  - flake8 .
  - ./.gofmt.sh
  - ./.buildifier.sh
- - bazel test reconciletags:reconciletags_test
  - bazel clean && bazel build //...
+ - bazel test reconciletags:reconciletags_test
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,5 +34,6 @@ script:
  - flake8 .
  - ./.gofmt.sh
  - ./.buildifier.sh
- - python -m unittest discover -s reconciletags -p 'reconciletags_test.py'
+ - bazel test reconciletags:reconciletags_test
+ - bazel test reconciletags:reconciletags_e2e_test
  - bazel clean && bazel build //...
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -32,8 +32,15 @@ new_go_repository(
 
 git_repository(
     name = "io_bazel_rules_docker",
+    commit = "db1b348dfdf161a784bc1efc5a1020395572b996",
     remote = "https://github.com/bazelbuild/rules_docker.git",
-    tag = "v0.0.1",
+
+)
+
+git_repository(
+    name = "containerregistry",
+    commit = "b0278a1544238d03648861b6d9395414d4c958e5",
+    remote = "https://github.com/google/containerregistry",
 )
 
 load(
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -48,3 +48,25 @@ load(
   "docker_repositories"
 )
 docker_repositories()
+
+new_http_archive(
+    name = "mock",
+    build_file_content = """
+# Rename mock.py to __init__.py
+genrule(
+    name = "rename",
+    srcs = ["mock.py"],
+    outs = ["__init__.py"],
+    cmd = "cat $< >$@",
+)
+py_library(
+   name = "mock",
+   srcs = [":__init__.py"],
+   visibility = ["//visibility:public"],
+)""",
+    sha256 = "b839dd2d9c117c701430c149956918a423a9863b48b09c90e30a6013e7d2f44f",
+    strip_prefix = "mock-1.0.1/",
+    type = "tar.gz",
+    url = "https://pypi.python.org/packages/source/m/mock/mock-1.0.1.tar.gz",
+)
+
--- a/reconciletags/BUILD
+++ b/reconciletags/BUILD
@@ -31,3 +31,15 @@ py_test(
     ],
 ) 
 
+py_test(
+    name = "reconciletags_e2e_test",
+    srcs = ["reconciletags_e2e_test.py",],
+    deps = [
+        "@containerregistry",
+        "@httplib2//:httplib2",
+    ],
+    data = [
+        "tiny_docker_image",
+    ],
+
+) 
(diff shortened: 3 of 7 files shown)


reconciletags/reconciletags.py
```diff
@@ -24,87 +24,92 @@
...
-            output = subprocess.check_output([command], shell=True)
-            logging.debug(output)
-            return output
+            src_name = docker_name.Digest(digest)
+            dest_name = docker_name.Tag(tag)
+            creds = docker_creds.DefaultKeychain.Resolve(src_name)
+            transport = transport_pool.Http(httplib2.Http)
+
+            with docker_image.FromRegistry(src_name, creds, transport) as src_img:
+                if src_img.exists():
```
dlorenc: What if it doesn't exist? Should we log and return an error?

reconciletags/reconciletags.py
```diff
@@ -24,87 +24,92 @@
...
+    def get_existing_tags(self, full_repo, digest):
+        full_digest = full_repo + '@sha256:' + digest
+        existing_tags = []
+
+        name = docker_name.Digest(full_digest)
+        creds = docker_creds.DefaultKeychain.Resolve(name)
+        transport = transport_pool.Http(httplib2.Http)
+
+        with docker_image.FromRegistry(name, creds, transport) as img:
+            if img.exists():
```
dlorenc: Same here.

reconciletags/reconciletags.py
```diff
@@ -24,87 +24,92 @@
...
-        command += " --format=" + fmt
+    def add_tags(self, digest, tag, dry_run): 
         if not dry_run:
-            logging.debug('Running {0}'.format(command))
-            output = subprocess.check_output([command], shell=True)
-            logging.debug(output)
-            return output
+            src_name = docker_name.Digest(digest)
+            dest_name = docker_name.Tag(tag)
+            creds = docker_creds.DefaultKeychain.Resolve(src_name)
```
dlorenc: This works in general, but I think in some cases we might need a different set of creds for src and dst. You should probably make another creds object below right before the push, based on the dest_name.

.travis.yml
```diff
@@ -34,5 +34,5 @@ script:
  - flake8 .
  - ./.gofmt.sh
  - ./.buildifier.sh
- - python -m unittest discover -s reconciletags -p 'reconciletags_test.py'
+ - bazel test reconciletags:reconciletags_test
```
dlorenc: Could you put this after the bazel clean && build line? It should be a bit faster :)

WORKSPACE
```diff
@@ -32,8 +32,15 @@ new_go_repository(
...
+
+)
+
+git_repository(
+    name = "containerregistry",
+    commit = "b0278a1544238d03648861b6d9395414d4c958e5",
+    remote = "https://github.com/google/containerregistry",
 )
 
 load(
```
dlorenc: Could you try pulling mock into the WORKSPACE via something like this:
https://github.com/bazelbuild/rules_k8s/pull/5/files#diff-09498dbadf45966909850dc8a47ebb13R46

reconciletags/BUILD
```diff
@@ -0,0 +1,41 @@
...
+    "go_prefix",
+)
+
+load(
+    "@io_bazel_rules_docker//docker:docker.bzl",
+    "docker_push",
+    "docker_build",
+)
+
+py_binary(
```
dlorenc: Can you add a "par_binary" target here as well, since you're loading in subpar/par_binary on line 3?

reconciletags/reconciletags.py
```diff
@@ -23,87 +23,99 @@
...
-                    self.add_tags(full_digest, full_tag, dry_run)
-
-                logging.debug(self.get_existing_tags(full_repo))
+                    name = docker_name.Digest(full_digest)
+                    creds = docker_creds.DefaultKeychain.Resolve(name)
+                    transport = transport_pool.Http(httplib2.Http)
+
+                    with docker_image.FromRegistry(
+                        name, creds, transport) as img:
+                            if img.exists():
```
dlorenc: To make this a little bit less nested, consider factoring out the logic for a single image into a function.

You can also do something like:

```python
if not img.exists():
  logging.warn()
  return
do other stuff
```

reconciletags/reconciletags.py
```diff
@@ -23,87 +23,99 @@
...
         return existing_tags
 
-    def get_latest_digest(self, repo):
-        output = json.loads(self.call('gcloud container images list-tags '
-                            '--no-show-occurrences {0}'.format(repo), False))
-        for image in output:
-            if 'latest' in image['tags']:
-                return image['digest']
+    def get_latest_digest(self, manifests):
+        for digest in manifests:
```
dlorenc: You could return an empty string at the end, in the other case where there is no latest. That will simplify the handling below.

reconciletags/reconciletags.py
```diff
@@ -23,87 +23,99 @@
...
+                        name, creds, transport) as img:
+                            if img.exists():
+                                existing_tags = img.tags()
+                                logging.debug("Existing Tags: {0}".format(existing_tags))
+
+                                manifests = img.manifests()
+                                latest = self.get_latest_digest(manifests)
+                                
+                                # Don't retag latest if it's already latest
+                                if latest:
```
dlorenc: If get_latest_digest always returns a string, you don't need to handle this case. I also think this is actually an existing bug, we don't want to special case 'latest' here. I think this logic instead should avoid retagging anything that already has that tag, not just for latest.

reconciletags/reconciletags_test.py
```diff
@@ -60,45 +64,80 @@ def setUp(self):
...
+            self.assertIn(self._tagging(_DIGEST1, _TAG1), logging_debug_output)
+            self.assertIn(_EXISTING_TAGS, logging_debug_output)
+
+    @patch('containerregistry.client.v2_2.docker_session.Push')
+    @patch('containerregistry.client.v2_2.docker_image.FromRegistry')
+    def test_dry_run(self, mock_from_registry, mock_push):
+        mock_from_registry.return_value = docker_image.FromRegistry()
+        mock_push.return_value = docker_session.Push()
+
+        with mock.patch('reconciletags.logging.debug') as mock_output:
```
dlorenc: You probably don't need to patch or have assertions on the logging calls, that's a bit overkill.

Sorry for not noticing this earlier!

reconciletags/reconciletags.py
sharifelgamal: Yep, this is definitely a bug. Thanks for catching this!

reconciletags/reconciletags_test.py
priyawadhwa: no worries! so it's enough to make sure the mocked functions are called?

PR #334, opened 2017-09-05
Update dependency versions

liyanhui1228:
Using the old versions caused the integration test build failure.


liyanhui1228:
@nkubala Please review it, thanks!

PR #340, opened 2017-09-08
Account for digest prefixes in retagger

priyawadhwa:


reconciletags/reconciletags.py
```diff
@@ -76,6 +76,21 @@ def get_tagged_digest(self, manifests, tag):
...
+
+        with docker_image.FromRegistry(name, creds, transport) as img:
+            digests = [d[len('sha256:'):] for d in img.manifests()]
+            matches = [d for d in digests if d.startswith(prefix)]
+            if len(matches) == 1:
+                return matches.pop()
+            elif len(matches) > 1:
+                logging.debug('{0} is not a unique digest prefix'.format(
+                                                                  prefix))
+        return ""
```
dlorenc: nit: single quotes to be consistent.

reconciletags/reconciletags.py
```diff
@@ -76,6 +76,21 @@ def get_tagged_digest(self, manifests, tag):
...
+    def get_digest(self, repo, prefix):
+        name = docker_name.Repository(repo)
+        creds = docker_creds.DefaultKeychain.Resolve(name)
+        transport = transport_pool.Http(httplib2.Http)
+
+        with docker_image.FromRegistry(name, creds, transport) as img:
+            digests = [d[len('sha256:'):] for d in img.manifests()]
+            matches = [d for d in digests if d.startswith(prefix)]
+            if len(matches) == 1:
+                return matches.pop()
```
dlorenc: nit: return matches[0] might be a bit more clear.

reconciletags/reconciletags.py
```diff
@@ -76,6 +76,21 @@ def get_tagged_digest(self, manifests, tag):
                 return digest
         return ""
 
+    def get_digest(self, repo, prefix):
```
dlorenc: get_digest_from_prefix would be a bit more explanatory

reconciletags/reconciletags_test.py
```diff
@@ -110,6 +110,22 @@ def test_add_tag(self, mock_from_registry, mock_push):
...
+    def test_get_digest(self, mock_from_registry):
+        fake_base = mock.MagicMock()
+        fake_base.manifests.return_value = ["sha256:" + _DIGEST1,
+                                            "sha256" + _DIGEST2]
+
+        mock_img = mock.MagicMock()
+        mock_img.__enter__.return_value = fake_base
+        mock_from_registry.return_value = mock_img
+
+        digest = self.r.get_digest(_FULL_REPO, "0")
```
dlorenc: Is "0" the first character in one of the _DIGEST constants? Maybe use it that way instead? _DIGEST[0:3] or something to be more clear it's a prefix.

A test case involving multiple matches would be good too.

priyawadhwa: yah, 0 is the first character, i'll make it clearer and add the test case!

reconciletags/reconciletags.py
```diff
@@ -75,7 +75,22 @@ def get_tagged_digest(self, manifests, tag):
...
+        name = docker_name.Repository(repo)
+        creds = docker_creds.DefaultKeychain.Resolve(name)
+        transport = transport_pool.Http(httplib2.Http)
+
+        with docker_image.FromRegistry(name, creds, transport) as img:
+            digests = [d[len('sha256:'):] for d in img.manifests()]
+            matches = [d for d in digests if d.startswith(prefix)]
+            if len(matches) == 1:
+                return matches[0]
+            elif len(matches) > 1:
```
dlorenc: This can just be an 'else', right?

reconciletags/reconciletags.py
```diff
@@ -88,7 +103,11 @@ def reconcile_tags(self, data, dry_run):
                 full_repo = os.path.join(registry, project['repository'])
 
                 for image in project['images']:
-                    full_digest = full_repo + '@sha256:' + image['digest']
+                    digest = self.get_digest_from_prefix(full_repo,
+                                                         image['digest'])
+                    if not digest:
+                        continue
```
dlorenc: I think we probably want to blow up and error here instead of continuing. WDYT?

priyawadhwa: I was thinking it could go ahead and tag other images if there were more in the json, but it's actually probably more useful to error

reconciletags/reconciletags.py
```diff
@@ -75,7 +75,22 @@ def get_tagged_digest(self, manifests, tag):
...
+        name = docker_name.Repository(repo)
+        creds = docker_creds.DefaultKeychain.Resolve(name)
+        transport = transport_pool.Http(httplib2.Http)
+
+        with docker_image.FromRegistry(name, creds, transport) as img:
+            digests = [d[len('sha256:'):] for d in img.manifests()]
+            matches = [d for d in digests if d.startswith(prefix)]
+            if len(matches) == 1:
+                return matches[0]
+            else:
```
sharifelgamal: what if len(matches) is zero? do we want a different error message for that?

PR #344, opened 2017-09-13
Test retagger par file in e2e test

priyawadhwa:


reconciletags/reconciletags_e2e_test.py
```diff
@@ -96,9 +97,32 @@ def testTagReconciler(self):
...
-        with self.assertRaises(subprocess.CalledProcessError):
+        with self.assertRaises(AssertionError):
             self.r.reconcile_tags(self._TEST_JSON, False)
 
+    def testParFile(self):
+        subprocess.call(['bazel', 'build', 'reconciletags:reconciletags.par'])
+        subprocess.call(['bazel-bin/reconciletags/reconciletags.par',
+                        'reconciletags/'+self._FILE_NAME])
+        # check list-tags to see if it added the correct tag
+        output = self._ListTags(self._REPO)
```
dlorenc: Could you factor out this with the section below?

reconciletags/reconciletags_e2e_test.py
```diff
@@ -96,9 +97,32 @@ def testTagReconciler(self):
 
         # now try with a fake digest
         self._TEST_JSON['projects'][0]['images'][0]['digest'] = 'fakedigest'
-        with self.assertRaises(subprocess.CalledProcessError):
+        with self.assertRaises(AssertionError):
             self.r.reconcile_tags(self._TEST_JSON, False)
 
+    def testParFile(self):
+        subprocess.call(['bazel', 'build', 'reconciletags:reconciletags.par'])
```
dlorenc: Could we make this whole test file run in bazel test? Then you won't need to do the build in here or hardcode the bazel-bin path below.

reconciletags/cloudbuild.yaml
```diff
@@ -0,0 +1,6 @@
+steps:
```
dlorenc: This is a bit of an unusual YAML, so it might be worth explaining what it does via comments. You might want to rename it to e2e_cloudbuild.yaml as well, to make it clear this is just used for e2e tests.

reconciletags/reconciletags_e2e_test.py
```diff
@@ -70,35 +70,42 @@ def setUp(self):
...
+    def checkListTagsOutput(self):
         output = self._ListTags(self._REPO)
         for image in output:
             if image['digest'].split(':')[1] == self.digest:
                 self.assertEquals(len(image['tags']), 2)
                 self.assertEquals(image['tags'][1], 'testing')
                 self.assertEquals(image['tags'][0], self._TAG)
 
-        # run reconciler again and make sure nothing changed
+    def testTagReconciler(self):
```
dlorenc: Does this test pass? This would have to be run through bazel I guess, right?

priyawadhwa: yah, it passes through bazel but not otherwsie!

PR #349, opened 2017-09-14
Package retagger into container with bazel

priyawadhwa:


reconciletags/BUILD
```diff
@@ -83,10 +83,24 @@ py_test(
...
 )
 
+docker_build(
+    name = "docker_credential_gcr",
+    base = "@python_base//image",
+    directory = "/usr/local/bin",
+    tars = ["@docker_credential_gcr//file"],
+)
+
+docker_build(
```
dlorenc: You can't leave a comment in the json file, but it might be worth explaining here in the BUILD file what this config is for. (it configures the client library to consult the credential helper for authenticating against GCR).

PR #352, opened 2017-09-19
Fixing publish builder so one malformed file doesn't break everyone

sharifelgamal:


runtime_builders/publish_builders.py
```diff
@@ -55,18 +55,25 @@ def main():
...
+            latest = config['latest']
+            project_name = config['project']
+
+            parts = os.path.splitext(latest)
+            prefix = builder_util.RUNTIME_BUCKET_PREFIX + project_name + '-'
+            latest_file = parts[0].replace(prefix, '')
+            file_name = '{0}.version'.format(project_name)
+            full_path = builder_util.RUNTIME_BUCKET_PREFIX + file_name
+            builder_util.write_to_gcs(full_path, latest_file)
+        except ValueError as ve:
```
dlorenc: What would throw the value error and the keyerror? Could we narrow the try/catch?

sharifelgamal: I'm not sure about the ValueError, I'll like take that out. The KeyError is for 

```python
latest = config['latest']
```

dlorenc: I'd replace the KeyError with an 'if 'latest' not in config or 'project' not in config'

sharifelgamal: PTAL

PR #354, opened 2017-09-21
Fixing up testing for new structure tests

sharifelgamal:


structure_tests/tests/debian_failure_test.yaml
```diff
@@ -0,0 +1,34 @@
+schemaVersion: '1.0.0'
```
dlorenc: Since this is yaml now, could you add some comments in here explaining what this is for?

structure_tests/tests/structure_test_tests.sh
```diff
@@ -20,43 +20,30 @@
...
-./build.sh "${STRUCTURE_TEST_IMAGE}"
-popd
+# build newest structure test binary
+go test -c .. -o structure-test
 
 # Run the debian tests, they should always pass on latest
-envsubst < cloudbuild.yaml.in > cloudbuild.yaml
-if ! gcloud container builds submit . --config=cloudbuild.yaml;
+res=$(./structure-test -image gcr.io/google-appengine/debian8 debian_test.yaml)
+code=$(echo $?)
```
dlorenc: do we need the echo here?

sharifelgamal: Nope, that's fixed.

structure_tests/tests/debian_failure_test.yaml
sharifelgamal: It's all in the README, but we really should be testing the newest schema version, so I'm going to fix that.

sharifelgamal:
These will run internally for now. If we find we need to run them in travis, we can set that up.

PR #357, opened 2017-09-22
Automate retagger releases

priyawadhwa:
This is the file the build trigger will execute


reconciletags_cloudbuild.yaml
```diff
@@ -0,0 +1,22 @@
+steps:
+#Build the par files with bazel
+  - name:  'gcr.io/cloud-builders/bazel'
+    args: ['build', 'reconciletags:reconciletags.par']
+  - name:  'gcr.io/cloud-builders/bazel'
+    args: ['build', 'reconciletags:config_integrity.par']
+  - name:  'gcr.io/cloud-builders/bazel'
+    args: ['build', 'reconciletags:data_integrity.par']
+#Copy par files into 'latest' bucket
```
dlorenc: Acrtually one last nit: could you push into the commit section first? In case something errors, I think we'd want the the commit one to be updated first.

dlorenc: And then you could make the second set of copies "server to server", so your src and dst would both be gs:// paths. That gets special-cased and is much faster since nothin needs to get uploaded.

dlorenc:
Do you still have to configure the triggers here? We can merge whenever you want, just let me know. I'm not sure what else needs to happen first so I'll leave it up to you.

PR #361, opened 2017-09-26
Fix godeps.

dlorenc:


dlorenc:
I had to rm -rf docgen to get this to work, then check it back out. I needed to manually update docker/docker and `go get github.com/containerd/continuity/pathdriver`. But then godep save worked and I can build again.

PR #372, opened 2017-10-04
fixing tests for structure tests

sharifelgamal:


sharifelgamal:
If only it did:

./structure_test_tests.sh: line 31: syntax error near `!=~'
./structure_test_tests.sh: line 31: `if [[ ("$res" !=~ "PASS" || "$code" != "0") ]];'

PR #373, opened 2017-10-09
python builder image with tests

aaron-prindle:


aaron-prindle:
PTAL

ftl/common/cache.py
```diff
@@ -99,3 +100,58 @@ def Store(self, base_image, namespace, checksum, value):
                 threads=self._threads,
                 mount=self._mount) as session:
             session.upload(value)
+
+
+class MockHybridRegistry(Base):
```
dlorenc: We already have this in test_util, any reason to have it here as well?

ftl/python/builder.py
```diff
@@ -0,0 +1,150 @@
...
+    def __enter__(self):
+        """Override."""
+        return self
+
+    def CreatePackageBase(self, base_image, cache, use_cache=True):
+        """Override."""
+        # Figure out if we need to override entrypoint.
+        # Save the overrides for later to avoid adding an extra layer.
+        descriptor = None
+        for f in [_REQUIREMENTS_TXT]:
```
dlorenc: You don't really need a loop since it's only one file.

aaron-prindle:
I think that is just because both files were empty

ftl/common/cache.py
aaron-prindle: no, removing

ftl/python/builder.py
```diff
@@ -0,0 +1,150 @@
...
+            logging.info('No package descriptor found. No packages installed.')
+        overrides = metadata.Overrides(
+            creation_time=str(datetime.date.today()) + "T00:00:00Z")
+        # TODO(aaron-prindle) currently if no requirements.txt, this fails here
+        checksum = hashlib.sha256(descriptor_contents).hexdigest()
+        if use_cache:
+            hit = cache.Get(base_image, _PYTHON_NAMESPACE, checksum)
+            if hit:
+                logging.info('Found cached dependency layer for %s' % checksum)
+                # TODO(aaron-prindle) check that cached dep layer is in TTL
```
dlorenc: Drop the TODO? Maybe make this a separate function though, so we can re-use it in Node.js

ftl/python/builder.py
```diff
@@ -0,0 +1,150 @@
...
+        logging.info('Starting venv creation ...')
+
+        # TODO(aaron-prindle) add support for different python versions
+        subprocess.check_call(
+            ['virtualenv', '--no-download', venv_dir, '-p', 'python3.6'],
+            cwd=app_dir)
+        os.environ['VIRTUAL_ENV'] = venv_dir
+        os.environ['PATH'] = venv_dir + "/bin" + ":" + os.environ['PATH']
+        my_env = os.environ.copy()
+        my_env.pop('PYTHONPATH', None)
```
dlorenc: A comment here would be good.

ftl/python/builder.py
```diff
@@ -0,0 +1,150 @@
...
+                cache.Store(base_image, _PYTHON_NAMESPACE, checksum, dep_image)
+            else:
+                logging.info('Skipping storing layer %s in cache.', sha)
+            return dep_image
+
+    def _gen_package_tar(self, descriptor, descriptor_contents):
+        tmp_app = tempfile.mkdtemp()
+        tmp_venv = tempfile.mkdtemp()
+
+        app_dir = os.path.join(tmp_app, 'app')
```
dlorenc: Can app_dir just be tmp_app? I don't think the name is actually important, like it is for venv.

aaron-prindle:
all comments resolved

ftl/python/builder.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+def From(ctx):
+    return Python(ctx)
+
+
+def _creation_time(image):
+    logging.info(image.config_file())
+    cfg = json.loads(image.config_file())
+    return cfg['created']
```
dlorenc: Can you use a .get here in case there's no cfg['created']

dlorenc:
I think after this goes in it would be a good time to refactor the ftl/common code to reduce duplication and boilerplate

ftl/python/builder.py

aaron-prindle:
Refactoring as a next step SGTM.  Definitely opportunities to dedupe for node and python

PR #375, opened 2017-10-10
Switch the structure test bazel rule to use the bazel built binary.

dlorenc:
cc @nkubala 

I got this working (not sure what the issue was before) and it seems to work for the debian-docker repo. I'll send a PR there after this goes in.


dlorenc:
Rebased.

PR #377, opened 2017-10-10
Testing library for containerregistry FromRegistry and session.Push

priyawadhwa:
Basic testing library for mocking FromRegistry and session.Push from the containerregistry library

### DIFF
--- a/testing/lib/mock_from_registry.py
+++ b/testing/lib/mock_from_registry.py
@@ -1,0 +1,47 @@
+# Copyright 2017 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from containerregistry.client.v2_2 import docker_image
+import mock_registry
+
+
+class MockFromRegistry(docker_image.FromRegistry):
+
+    REGISTRY = mock_registry.MockRegistry()
+
+    def __init__(self, name="", basic_creds=None, transport=None):
+        self._name = name
+
+    def setRegistry(self, registry):
+        if isinstance(registry, mock_registry.MockRegistry):
+            MockFromRegistry.REGISTRY = registry
+
+    def exists(self):
+        return MockFromRegistry.REGISTRY.existsImage(self._name)
+
+    def manifests(self):
+        return MockFromRegistry.REGISTRY.getManifests(self._name)
+
+    def tags(self):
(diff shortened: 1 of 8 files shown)


testing/lib/mock_from_registry.py
```diff
@@ -0,0 +1,47 @@
...
+class MockFromRegistry(docker_image.FromRegistry):
+
+    REGISTRY = mock_registry.MockRegistry()
+
+    def __init__(self, name="", basic_creds=None, transport=None):
+        self._name = name
+
+    def setRegistry(self, registry):
+        if isinstance(registry, mock_registry.MockRegistry):
+            MockFromRegistry.REGISTRY = registry
```
dlorenc: I think this is setting this as a class variable, not on the instance itself. Should you set this as self. registry instead of MockFromRegistry.REGISTRY?

testing/lib/mock_registry.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def getRegistry(self):
+        return self
+
+    def getFullRepoStr(self, full_repo):
+        if isinstance(full_repo, docker_image.FromRegistry):
+            full_repo = full_repo.getName()
+        return str(full_repo)
+
+    def getRepoStr(self, repository):
```
dlorenc: Can you re-use any of the existing __str__ methods in docker_name.py here?

testing/lib/mock_from_registry.py
priyawadhwa: I set it as a class variable so that if an image was pushed from session.Push then FromRegistry could still access it. I wasn't sure if there was a better way to do this?

priyawadhwa:
Hey @dlorenc, sorry but what were you referring to when you said "In other places this is a dictionary, not a list"?

testing/lib/mock_from_registry.py
dlorenc: Ah I see - I think in that case it should probably be up to the "WithMockRegistry" class to create a MockRegistry instance and store it.

Does that make sense?

dlorenc:
> Hey @dlorenc, sorry but what were you referring to when you said "In other places this is a dictionary, not a list"?

Weird - something strange is happening with my github comments. I think I tried to delete that one but it showed up as detached or something. Don't worry about it.

testing/lib/mock_from_registry.py
priyawadhwa: I think so, is that the same as what I was doing in the tests [here](https://github.com/priyawadhwa/runtimes-common/blob/c84ddfdc69d725a9dcb7bdac4cacdc196c430a84/testing/lib/mock_registry_tests.py#L30)? I can move it into "WithMockRegistry" if so.

dlorenc: Yup!

priyawadhwa: I ended up moving it into a separate file for clarity!

testing/lib/mock_registry.py
priyawadhwa: I looked, but couldn't find anything :(

Systemreiniger092016:
Whats is about testing AndroidCloudPlatform:user 
to certificate a mobile device may a tablet with no Android Certificate in Europa?

I Wish hopefully to helping me in step by step order like beginner.

Thanksgiving you so much ☺

PR #378, opened 2017-10-11
Make structure tests work with symlinks.

dlorenc:


dlorenc:
This should fix the issue @aaron-prindle hit in https://github.com/GoogleCloudPlatform/debian-docker/pull/98

PR #383, opened 2017-10-13
Stop using some context managers that we don't need.

dlorenc:
This makes it a bit easier to read IMO.


dlorenc:
cc @aaron-prindle

PR #389, opened 2017-10-16
Write main test for node

priyawadhwa:


.travis.yml
```diff
@@ -42,6 +42,7 @@ script:
  # - bazel clean && bazel build //...
  - bazel test reconciletags:reconciletags_test
  - bazel run //ftl:node_builder_test
```
dlorenc: I don't think this one needs a "bazel run", could you combine these two to "bazel test //ftl:..."

ftl/node/main_test.py
```diff
@@ -0,0 +1,51 @@
...
+                                       ) as img:
+            self.registry.setImage('fake.gcr.io/base/image:appended', img)
+
+        self.AssertPushed(self.registry, 'fake.gcr.io/base/image:initial')
+        self.AssertPushed(self.registry, 'fake.gcr.io/base/image:appended')
+
+        append_layer_mock.return_value = self.registry.getImage(
+                                         'fake.gcr.io/base/image:appended')
+
+        sys.argv = ["main.py",  "--base=fake.gcr.io/base/image:initial",
```
dlorenc: Could you make a slight refactor to main to take these as parameters? Then you can pass it in instead of setting sys.argv.

aaron-prindle:
Is main_test.py going going to house all of the integration tests for ftl node?  What is the primary functionality intended to be tested for main.py in main_test.py?

priyawadhwa:
Right now main_test.py primarily tests that the image is uploaded correctly, but I'm not sure if it's where all of the integration tests are going to be...

.travis.yml
```diff
@@ -41,7 +41,7 @@ script:
  - ./.buildifier.sh
  # - bazel clean && bazel build //...
  - bazel test reconciletags:reconciletags_test
- - bazel run //ftl:node_builder_test
+ - bazel test ftl:node_builder_test ftl:node_main_test
  - bazel test testing/lib:mock_registry_tests
```
dlorenc: could you make this :... to test all the files in this package?

ftl/node/main.py
```diff
@@ -47,9 +47,16 @@
     help='The path where the application data sits.')
 
 
-def main():
+def main(base="", name="", directory=""):
```
dlorenc: Ah sorry, I wasn't very clear here. Could you pass sys.argv to main from the bottom of this file, and pass that parameter to parser.parse_args?

ftl/node/main.py
```diff
@@ -90,4 +91,4 @@ def main():
 
 
 if __name__ == '__main__':
-    main()
+    main(sys.argv)
```
dlorenc: One more tweak, this is a bit easier to use if you pass argv[1:] here, then you don't need the "main.py" in your test.

priyawadhwa: This all makes a lot more sense, thanks Dan!

PR #396, opened 2017-10-18
added -v level logging to ftl/node/main.py

aaron-prindle:


ftl/node/main.py
```diff
@@ -46,9 +46,14 @@
     action='store',
     help='The path where the application data sits.')
 
+parser.add_argument("-v", "--verbose",
+    const=1, default=0, type=int, nargs="?",
+    help="increase verbosity: 0 = only warnings, 1 = info, 2 = debug. " +
```
dlorenc: I kind of prefer using the actual strings (debug, warning, info) etc. as flags. WDYT?

aaron-prindle: sgtm

aaron-prindle:
updated logging to use syntax:
main.py --v (DEBUG, INFO, etc.)

ftl/node/main.py
```diff
@@ -46,9 +46,14 @@
     action='store',
     help='The path where the application data sits.')
 
+parser.add_argument("-v", "--verbose",
```
dlorenc: nit: --verbosity as the long form?

dlorenc: And variable name. You can use the "choices" parameter here to get the help text and validation for free.

aaron-prindle: both done

dlorenc:
Looks like a flake issue.

aaron-prindle:
fixed

PR #398, opened 2017-10-19
updated base image for ft/node to use debian-docker ubuntu img

aaron-prindle:


aaron-prindle:
this PR is to use the bazel nodejs+ubuntu+build-essential/python2.7 image from debian-docker as the base in our FTL image

ftl/BUILD
```diff
@@ -100,13 +91,22 @@ py_binary(
...
     srcs = [":node_lib"],
-    base = ":node_builder_base",
+    base = "@debian_docker//reproducible/ubuntu:nodejs_ubuntu_build",
     main = "node/main.py",
     deps = [
         "@containerregistry",
     ],
 )
 
+# The node base image sets a default CMD, which causes issues if you run our image with no CMD.
```
dlorenc: I think you can actually just delete this rule since you're not using it anymore.

aaron-prindle: its used for testing

aaron-prindle: https://github.com/aaron-prindle/runtimes-common/blob/ca9a0946b9890967ac4b0d2ef5b343c9f0dc1f48/ftl/BUILD#L113

PR #401, opened 2017-10-19
support gcp-build script

priyawadhwa:


ftl/node/builder.py
```diff
@@ -63,13 +64,14 @@ def CreatePackageBase(self, base_image, cache):
...
-
         # Copy out the relevant package descriptors to a tempdir.
         with open(os.path.join(app_dir, descriptor), 'w') as f:
             f.write(self._ctx.GetFile(descriptor))
 
         tar_path = tempfile.mktemp()
-        subprocess.check_call(['npm', 'install', '--no-cache'], cwd=app_dir)
+        check_gcp_build(json.loads(self._ctx.GetFile(_PACKAGE_JSON)), app_dir)
+        subprocess.check_call(['rm', '-rf', 'node_modules'], cwd=app_dir)
+        subprocess.check_call(['npm', 'install', '--production', '--no-cache'])
```
dlorenc: I think you need the cwd app_dir here.

ftl/node/builder.py
```diff
@@ -63,13 +64,14 @@ def CreatePackageBase(self, base_image, cache):
...
         os.mkdir(app_dir)
-
         # Copy out the relevant package descriptors to a tempdir.
         with open(os.path.join(app_dir, descriptor), 'w') as f:
             f.write(self._ctx.GetFile(descriptor))
 
         tar_path = tempfile.mktemp()
-        subprocess.check_call(['npm', 'install', '--no-cache'], cwd=app_dir)
+        check_gcp_build(json.loads(self._ctx.GetFile(_PACKAGE_JSON)), app_dir)
+        subprocess.check_call(['rm', '-rf', 'node_modules'], cwd=app_dir)
```
dlorenc: Probably don't need the cwd if you os.path.join(app_dir, 'node_modules')

ftl/node/builder.py
```diff
@@ -85,5 +90,17 @@ def CreatePackageBase(self, base_image, cache):
...
 
 
+def check_gcp_build(package_json, app_dir):
+    scripts = package_json.get('scripts', {})
+    gcp_build = scripts.get('gcp-build')
+
+    if not gcp_build:
+        return
+
+    os.environ["NODE_ENV"] = "development"
```
dlorenc: We probably need to drop the os.environ after we finish these two. You could either grab the current environment as a dictionary, set NODE_ENV, and pass it into check_call (so it only applies to these calls), or add a try: finally clause here to make sure it gets unset after.

PR #408, opened 2017-10-25
kokoro presubmit for ftl integration test

priyawadhwa:


dlorenc:
Looks like you need a shellcheck disable.

dlorenc:
Nice! although it doesn't look like kokoro actually ran..

priyawadhwa:
Yah, I think the CL has to go through first and then hopefully it'll work!

priyawadhwa:
kokoro:run

dlorenc:
Looks like the kokoro service account is missing permissions to push to ftl-node-test.

PR #411, opened 2017-10-26
Add support for node.js entrypoints to the FTL builder.

dlorenc:


ftl/node/builder.py
```diff
@@ -106,3 +124,17 @@ def check_gcp_build(package_json, app_dir):
...
+
+    scripts = package_json.get('scripts', {})
+    start = scripts.get('start', _DEFAULT_ENTRYPOINT)
+    prestart = scripts.get('prestart')
+
+    if prestart:
+        entrypoint = '%s && %s' % (prestart, start)
+    else:
+        entrypoint = start
+    return ['sh', '-c', "'%s'" % entrypoint]
```
aaron-prindle: nit:  I believe the environments that 'sh -c <command>' and <command> run in are a tiny bit different (subshell vs parent shell).  It might be worth making the calls both use 'sh -c' or neither use it just to make sure the env's are the same w/ or w/o prestart

PR #413, opened 2017-10-30
revert cloudbuild rule change for ftl integration test

aaron-prindle:

PR #414, opened 2017-10-30
Update ftl integration tests to use bazel run

priyawadhwa:
Added --image flag so that bazel run can build the image without errors, so that it can be used as a build step in the ftl integration tests.

#413

### DIFF
--- a/ftl/ftl_node_integration_tests.yaml
+++ b/ftl/ftl_node_integration_tests.yaml
@@ -1,7 +1,7 @@
 steps:
 #Build node ftl image with bazel
   - name:  'gcr.io/cloud-builders/bazel'
-    args: ['run', '//ftl:node_builder_image.tar']
+    args: ['run', '--', '//ftl:node_builder_image', '--image']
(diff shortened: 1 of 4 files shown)


ftl/ftl_node_integration_tests.yaml
```diff
@@ -1,7 +1,7 @@
 steps:
 #Build node ftl image with bazel
   - name:  'gcr.io/cloud-builders/bazel'
```
dlorenc: What's the issue here? That doing "bazel run" on this image tries to load it *and* run it, which errors? Run should just do the "load", but if it's actually doing a run as well I think we can handle that here by setting `legacy_run_behavior` to True:
https://github.com/bazelbuild/rules_docker/blob/master/container/image.bzl#L317

priyawadhwa: Yah, so the "bazel run" does seem to be loading and running the image, I'll try using legacy_run_behavior to fix it!

PR #416, opened 2017-11-01
Added cloudbuild rule ftl_node_release.yaml and build trigger

aaron-prindle:


aaron-prindle:
This will fail until:
https://github.com/GoogleCloudPlatform/debian-docker/pull/123
goes in.  Then I will update the commit hash for `debian-docker`

PR #417, opened 2017-11-01
Move retagger code and tests into single binary

priyawadhwa:


reconciletags/reconciletags_main.py
```diff
@@ -0,0 +1,99 @@
...
+    os.makedirs(dir)
+
+    for file in files:
+        if os.path.isfile(file):
+            shutil.copy(file, dir)
+        else:
+            raise AssertionError("{0} is not a valid file".format(file))
+
+
+def delete_temp_dir():
```
dlorenc: This scares me a bit, can we make whatever is creating this use a real tempdir instead of a path in the repo?

reconciletags/reconciletags_main.py
```diff
@@ -0,0 +1,99 @@
...
+from reconciletags import tag_reconciler
+from reconciletags import data_integrity_test
+from reconciletags import config_integrity_test
+
+
+def create_temp_dir(files):
+    dir = "../config/tag/"
+    os.makedirs(dir)
+
+    for file in files:
```
dlorenc: I think you can use shutil copytree here.

priyawadhwa: I think that might not work since the files could be in different directories

priyawadhwa:
No, I think the only documentation was going to be the blog post

PR #418, opened 2017-11-01
Unquote the entrypoint argument.

mattmoor:
e.g. 'sh', '-c', 'over-quoted'


aaron-prindle:
It seems the test checks for output with the quotes, I think you'll need to change the expected values there too:
https://travis-ci.org/GoogleCloudPlatform/runtimes-common/builds/296022320#L1705
https://github.com/GoogleCloudPlatform/runtimes-common/blob/master/ftl/node/builder_test.py#L149

PR #421, opened 2017-11-02
added additional logging to ftl for benchmarking

aaron-prindle:


ftl/common/builder.py
```diff
@@ -73,17 +73,21 @@ def CreatePackageBase(self, base_image, cache):
...
                 info = tarfile.TarInfo(os.path.join('app', name))
                 info.size = len(content)
                 out.addfile(info, fileobj=cStringIO.StringIO(content))
+        logging.info('Finished generating tarfile from context.')
 
         tar = buf.getvalue()
         sha = 'sha256:' + hashlib.sha256(tar).hexdigest()
 
         gz = cStringIO.StringIO()
+        logging.info('Starting to gzip tarfile...')
```
dlorenc: Do these include timestamps by default when printed, or do we need to customize the formatter?

aaron-prindle: no it didn't.  I just updated the logging configuration to have millisecond timings so now it does.

dlorenc:
Looks like there's a flake issue with line length.

aaron-prindle:
passing now

PR #422, opened 2017-11-02
Add local mode to ftl

priyawadhwa:


ftl/node/main.py
```diff
@@ -96,6 +108,18 @@ def main(args):
             logging.info('Generating app layer...')
             app_layer, diff_id = bldr.BuildAppLayer()
             with append.Layer(deps, app_layer, diff_id=diff_id) as app_image:
+                if args.local:
+                    if not args.filepath:
+                        raise AssertionError('Please provide path to tar \
+                                             with --filepath flag')
+                    dir = os.path.dirname(args.filepath)
```
dlorenc: You could leave off this check and just assume the user created a tar ahead of time.

dlorenc: We might also want to swap out or disable the cache if --local mode is used. WDYT?

priyawadhwa: I think disabling it could make sense, but what would we swap it out for?

priyawadhwa:
Hey @dlorenc, do we actually need the --local flag? It could be implied we want to save it locally if the --filepath is provided

dlorenc:
> Hey @dlorenc, do we actually need the --local flag? It could be implied we want to save it locally if the --filepath is provided

Nope! Good call.

ftl/node/main.py
```diff
@@ -56,6 +57,12 @@
     help='The path where the application data sits.')
 
 parser.add_argument(
+    '--filepath',
```
dlorenc: Nit: Something like --output-path might be a little more clear.

PR #424, opened 2017-11-03
Add shellcheck and other deps in kokoro.

dlorenc:


dlorenc:
Small test refactor as well to avoid needing NPM.

dlorenc:
OK, this should actually be ready now and will fix kokoro.

PR #425, opened 2017-11-03
Ftl no cache flag

priyawadhwa:


ftl/node/main.py
```diff
@@ -57,6 +57,12 @@
     help='The path where the application data sits.')
 
 parser.add_argument(
+    '--no-cache',
```
dlorenc: double negatives can be a bit hard to reason about, it might make sense to name this flag --cache and default it to true. Then we don't need "if not no_cache" above.

ftl/node/builder.py
```diff
@@ -66,12 +66,13 @@ def CreatePackageBase(self, base_image, cache):
                 base_image, tar_gz=None, overrides=overrides)
 
         checksum = hashlib.sha256(descriptor_contents).hexdigest()
-        hit = cache.Get(base_image, _NODE_NAMESPACE, checksum)
-        if hit:
-            logging.info('Found cached dependency layer for %s' % checksum)
-            return hit
-        else:
-            logging.info('No cached dependency layer for %s' % checksum)
+        if not no_cache:
```
dlorenc: Can we log whenever we skip the cache check?

ftl/node/main.py
priyawadhwa: so then when --cache is passed in, that actually means not to use the cache? would it make more sense to keep it as --no-cache, and pass the negative in to the builder?

dlorenc: Sorry, to be clear --cache would be a boolean that defaults to true. So you would do --cache=False to turn it off.

ftl/node/main.py
```diff
@@ -99,7 +105,8 @@ def main(args):
         # Create (or pull from cache) the base image with the
         # package descriptor installation overlaid.
         logging.info('Generating dependency layer...')
-        with bldr.CreatePackageBase(base_image, cash) as deps:
+        with bldr.CreatePackageBase(base_image, cash,
+                                    eval(args.cache)) as deps:
```
dlorenc: Why do we need the "eval" here? Would a "bool" work?

priyawadhwa: So the flag gets passed in a string, and the bool of a non-empty string will always be True, so I used eval instead!

dlorenc:
Ah, it looks like this is actually the correct way to do it:

```python
parser.add_argument('--feature', dest='feature', action='store_true')
parser.add_argument('--no-feature', dest='feature', action='store_false')
```

from here: https://stackoverflow.com/questions/15008758/parsing-boolean-values-with-argparse

PR #429, opened 2017-11-03
Merge the two presubmits into one.

dlorenc:


dlorenc:
cc @sharifelgamal there's a pending change in the internal config to drop this other job as well.

dlorenc:
> LGTM. Should this presubmit also be used as a pre-commit hook?

It should essentially be the same (test.sh is called in the pre-commit hook). The only difference should be integration tests and pre-req installation.

PR #434, opened 2017-11-09
Push a .par binary to GCS in releases.

dlorenc:


ftl/ftl_node_release.yaml
```diff
@@ -1,6 +1,10 @@
 steps:
-  - name:  'gcr.io/cloud-builders/bazel'
+  - name: 'gcr.io/cloud-builders/bazel'
     args: ['run','//ftl:node_builder_image', '--', '--norun']
   - name: gcr.io/cloud-builders/docker:latest
     args: ['tag', 'bazel/ftl:node_builder_image', 'gcr.io/gcp-runtimes/ftl_node_8_5_0_ubuntu_16_0_4_build:${COMMIT_SHA}']
+  - name: 'gcr.io/cloud-builders/bazel'
+    args: ['build','//ftl:node_builder.par']
+  - name: 'gcr.io/cloud-builders/gsutil'
+    args: ['cp', 'bazel-bin/ftl/node_builder.par', 'gs://gcp-container-tools/ftl/node/${COMMIT_SHA}/ftl.par']
```
aaron-prindle: Does the the ftl.par filename need to reference the node or ubuntu version?

dlorenc: No, the idea is that this par can be dropped into an image at a set version of node/ubuntu. It's agnostic by itself.

aaron-prindle: Got it, makes sense.

PR #436, opened 2017-11-09
Add an integration test for the package-lock and package.json case.

dlorenc:
I expect this to fail, then I'll add a commit to make the new test pass.


dlorenc:
This is RFAL. I have another PR on the way to refactor the way we add integration tests to make it a bit easier.

PR #437, opened 2017-11-09
Add a cache version we can use to globally bust caches.

dlorenc:


ftl/node/main.py
```diff
@@ -94,6 +94,10 @@
     choices=_LEVEL_MAP.keys())
 
 
+# Version string used to bust caches.
+_NODE_CACHE_VERSION = 'v1'
```
aaron-prindle: would this make more sense as an arg to the command?  If I understand correctly, we would need to change the source to bust the cache.  Is this desired?

dlorenc: Yeah that's basically the idea, we'd bump this in the source before cutting a release that is backwards-incompatible.

aaron-prindle: SGTM

PR #439, opened 2017-11-09
added destination flag to specify base path for node_modules dir

aaron-prindle:


ftl/node/main.py
```diff
@@ -86,6 +85,14 @@
           instead of pushing to registry')
 
 parser.add_argument(
+    '--destination',
+    dest='destination_path',
+    action='store',
+    default="app",
```
dlorenc: Should this be /app? Can you make the quoting consistent as well?

aaron-prindle: done

aaron-prindle:
integration test added

ftl/node/builder.py
```diff
@@ -73,29 +77,29 @@ def CreatePackageBase(self, base_image, cache, use_cache=True):
...
+        # We want the node_modules directory rooted at:
+        # $destination_path/node_modules in the final image.
         # So we build a hierarchy like:
-        # /$tmp/app/node_modules
+        # /$tmp/$destination_path/node_modules
         # And use the -C flag to tar to root the tarball at /$tmp.
         tmp = tempfile.mkdtemp()
-        app_dir = os.path.join(tmp, 'app')
-        os.mkdir(app_dir)
+        app_dir = os.path.join(tmp, destination_path.strip("/"))
```
dlorenc: Do you need strip with os.path.join?

ftl/ftl_node_integration_tests_yaml.py
```diff
@@ -30,11 +32,12 @@ def main():
...
 
     # Generate a set of steps for each test and add them.
-    for test in ['gcp_build_test', 'packages_test', 'packages_lock_test']:
-        cloudbuild_yaml['steps'] += run_test_steps(
-            _NODE_BASE,
-            '%s-image' % test,
-            test)
+    for test in [
+            'gcp_build_test', 'packages_test', 'packages_lock_test',
+            'destination_test'
```
dlorenc: This loop won't quite work, since you need special args for destination-app passed to ftl:node_builder_image.

ftl/node/builder.py
aaron-prindle: I believe so from:
'''
aprindle@aprindle-glaptop0:~$ python
Python 2.7.6 (default, Oct 26 2016, 20:30:19) 
[GCC 4.8.4] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.path.join("test", "test")
'test/test'
>>> os.path.join("test", "/test")
'/test'
'''

PR #441, opened 2017-11-13
Ftl benchmarking

priyawadhwa:
To help benchmark FTL, the main file runs the node_builder binary 10 times on a node app and pushes the data to a bigquery table. I run this on three node apps, of small/medium/large sizes, through container builder.


ftl/node/benchmark/small_app.yaml
```diff
@@ -0,0 +1,17 @@
+# This is a 'small' node app, with 1 dependency (20th percentile of all node projects with dependencies)
+steps:
+# Build benchmark image with bazel
+  - name:  'gcr.io/cloud-builders/bazel'
+    args: ['run','ftl/node/benchmark:node_benchmark_image', '--', '--norun']
+# Clone git repo
+  - name:  'gcr.io/cloud-builders/git'
+    args: ['clone','https://github.com/KevinGaudin/monsterid.js.git']
```
aaron-prindle: We should probably host the sample apps so that there aren't unexpected changes in the app, for licensing purposes, and also in case the repo is deleted.

priyawadhwa: oh good point, where should we host them?

dlorenc: Yeah it would be easier if we can find ones we already host with the same number of deps, then commit just the package.jsons to a directory in this repo

ftl/node/benchmark/main.py
```diff
@@ -0,0 +1,95 @@
...
+import datetime
+import time
+import os
+import logging
+from google.cloud import bigquery
+
+DATASET_NAME = 'ftl_benchmark'
+TABLE_NAME = 'ftl_benchmark'
+PROJECT_NAME = 'ftl-node-test'
+NUM_ITERATIONS = 10
```
dlorenc: It might make sense to make this one a flag.

ftl/node/benchmark/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM gcr.io/google-appengine/python
+RUN pip install google-cloud
+RUN apt-get update
```
dlorenc: These three lines should all be on the same line: RUN apt-get update && apt-get -y install nodejs npm && apt-get clean && \
    rm /var/lib/apt/lists/*_*

(from here: https://github.com/GoogleCloudPlatform/base-images-docker/blob/35e20794b1770f3bb7e585e0b33f8bdab6827643/ubuntu/Dockerfile.ubuntu)

Although we probably want a specfic version of nodejs/npm, so it might be easier to use the ubuntu/nodejs image we build over here: https://github.com/GoogleCloudPlatform/runtimes-common/blob/master/ftl/BUILD#L100

It should already have python.

dlorenc:
Just thought of this and I'm not sure how practical it is, but would it be possible to store the raw stdout/stderr of hte FTL subprocess in bigquery? It contains some detailed timings we could use later on.

priyawadhwa:
I think it should be possible, I can try it and see what it looks like!

dlorenc:
So the current failure is due to a bazel cache issue. It's probably easiest if you change test.sh to exclude the benchmark image from being built during presubmit tests. WDYT?

priyawadhwa:
SGTM!

PR #445, opened 2017-11-14
ftl refactor to dedupe runtimes

aaron-prindle:


ftl/common/builder_runner.py
```diff
@@ -0,0 +1,97 @@
...
+        self.cache_version_str = cache_version_str
+
+        self._init_transport()
+        self._init_docker_config(args)
+        self._init_ctx(args)
+        self._init_cache(cache_version_str)
+        self.bldr = builder.From(self.ctx)
+
+    def _init_transport(self):
+        self.transport = transport_pool.Http(httplib2.Http, size=_THREADS)
```
dlorenc: You can do most of this right in the constructor instead of breaking them out.

ftl/common/builder_runner.py
```diff
@@ -0,0 +1,97 @@
...
+            mount=[self.base_name])
+
+    def GenerateFTLImage(self):
+        with docker_image.FromRegistry(self.base_name, self.base_creds,
+                                       self.transport) as base_image:
+
+            # Create (or pull from cache) the base image with the
+            # package descriptor installation overlaid.
+            logging.info('Generating dependency layer...')
+            with self.bldr.CreatePackageBase(base_image, self.cash,
```
dlorenc: It seems like we could extract the cache getting/setting as well by exposing some methods on builder around the cache key generation. WDYT?

aaron-prindle:
PTAL

PR #448, opened 2017-11-17
PHP FTL

priyawadhwa:


ftl/php/builder.py
```diff
@@ -0,0 +1,93 @@
...
+
+        # Copy out the relevant package descriptors to a tempdir.
+        for f in self.descriptor_files:
+            if self._ctx.Contains(f):
+                with open(os.path.join(app_dir, f), 'w') as w:
+                    w.write(self._ctx.GetFile(f))
+
+        tar_path = tempfile.mktemp()
+        logging.info('Starting composer install ...')
+        subprocess.check_call(
```
squee1945: Aside: in my manual testing of `composer install` so far, I've found it to be very slow (as compared to `npm install`). Any suggestions are appreciated!

ftl/php/builder.py
```diff
@@ -0,0 +1,93 @@
...
+        subprocess.check_call(['tar', '-C', tmp, '-cf', tar_path, '.'])
+        logging.info('Finished generating tarfile for composer packages.')
+
+        # We need the sha of the unzipped and zipped tarball.
+        # So for performance, tar, sha, zip, sha.
+        # We use gzip for performance instead of python's zip.
+        sha = 'sha256:' + hashlib.sha256(open(tar_path).read()).hexdigest()
+
+        logging.info('Starting to gzip composer package tarfile...')
+        subprocess.check_call(['gzip', tar_path])
```
squee1945: Should this be a 0-compression gzip (basically a no-op)?

aaron-prindle: yea we should probably drop the compression level to 0 or 1 here to speedup the run.  We can experiment with the values using our benchmarks and find the best balance for the cpu/network tradeoff.

hack/kokoro-test.sh
```diff
@@ -25,3 +25,4 @@ echo "Running integration tests..."
 # Generate the integration test yaml and pass it to cloud build via stdin.
 python ftl/ftl_node_integration_tests_yaml.py | gcloud container builds submit --config /dev/fd/0 .
 gcloud container builds submit --config ftl/ftl_python_integration_tests.yaml .
+gcloud container builds submit --config ftl/ftl_php_integration_tests.yaml .
```
aaron-prindle: can you change this to a yaml test generator like the node tests?  That way it'll be easier to extend in the future and we can dedupe some of test logic

ftl/php/builder.py
priyawadhwa: I ran the benchmarks, and on average 0 was faster than 1 by almost 3 seconds, so I'll go with that!

ftl/php/builder.py
```diff
@@ -0,0 +1,93 @@
...
+        subprocess.check_call(['tar', '-C', tmp, '-cf', tar_path, '.'])
+        logging.info('Finished generating tarfile for composer packages.')
+
+        # We need the sha of the unzipped and zipped tarball.
+        # So for performance, tar, sha, zip, sha.
+        # We use gzip for performance instead of python's zip.
+        sha = 'sha256:' + hashlib.sha256(open(tar_path).read()).hexdigest()
+
+        logging.info('Starting to gzip composer package tarfile...')
+        subprocess.check_call(['gzip', tar_path, '-1'])
```
dlorenc: Could you add a comment about the compression level?

ftl/php/builder.py
```diff
@@ -0,0 +1,93 @@
...
+from ftl.common import builder
+
+_PHP_NAMESPACE = 'php-composer-lock-cache'
+_COMPOSER_LOCK = 'composer.lock'
+_COMPOSER_JSON = 'composer.json'
+
+
+class PHP(builder.JustApp):
+    def __init__(self, ctx):
+        self.descriptor_files = [_COMPOSER_LOCK, _COMPOSER_JSON]
```
dlorenc: Could this be a constant art the package level? We never change it afaict.

ftl/php/builder.py
```diff
@@ -0,0 +1,93 @@
...
+        self.namespace = _PHP_NAMESPACE
```
dlorenc: We shouldn't need this on self since we never change it.

priyawadhwa: I think we need it because it's called in builder_runner [here](https://github.com/GoogleCloudPlatform/runtimes-common/blob/master/ftl/common/builder_runner.py#L75).

ftl/php/builder.py
priyawadhwa: I think same as with the namespace, it's called [here](https://github.com/GoogleCloudPlatform/runtimes-common/blob/master/ftl/common/builder_runner.py#L105) so it needs to be on self?

ftl/php/builder.py
dlorenc: Ah sorry, missed that in the recent refactor.

PR #453, opened 2017-11-20
Add an FTL doc section.

dlorenc:


ftl/README.md
```diff
@@ -1,4 +1,26 @@
...
+```
+
+This command can be read as "Build the source code in directory `$dir` into an image named `$img`, based on the image `$base`.
+
+These binaries **do not depend on Docker**, and construct images directly in the registry.
+
+## Developing
 To run the FTL integration tests, run the following command locally from the root directory:
 ```
 gcloud container builds submit --config ftl/ftl_node_integration_tests.yaml .
```
priyawadhwa: I don't think this file exists anymore, it was changed to be created by ftl_node_integration_tests_yaml.py

PR #462, opened 2017-11-21
phase1.5 implementation for ftl python

aaron-prindle:


dlorenc:
```shell
Step #5: --- FAIL: TestAll (0.06s)
Step #5:     --- FAIL: TestAll/File_Existence_Test:_flask_lib (0.06s)
Step #5:     	structure.go:94: File /env/lib/python3.6/site-packages/flask should exist but does not, got error: API error (404): {"message":"lstat /var/lib/docker/aufs/mnt/42ff46f71f2954f0035e54e2dbdd095fed62cbed7b781ceb4283e2e4682ed288/env/lib/python3.6/site-packages/flask: no such file or directory"}
Step #5:     		!
Step #5: 	structure_test.go:46: Total tests run: 2
Step #5: FAIL
```

aaron-prindle:
This is because travis is having issues using 3.6, changing to 2.7 but didn't change tests.  Doing so now

aaron-prindle:
Will add --python-version arg once arg passing is refactored

aaron-prindle:
PTAL

ftl/python/builder.py
```diff
@@ -33,76 +33,106 @@ class Python(builder.JustApp):
...
-    def _generate_overrides(self):
+    def _generate_overrides(self, py_version='python2.7'):
         return metadata.Overrides(
-            creation_time=str(datetime.date.today()) + "T00:00:00Z")
-
-    def CreatePackageBase(self, base_image, cache):
+            creation_time=str(datetime.date.today()) + "T00:00:00Z",
+            env={
+                "VIRTUAL_ENV":
+                "/env",  # not working properly, not sure why
```
dlorenc: What's not working about this?

ftl/python/builder.py
```diff
@@ -33,76 +33,106 @@ class Python(builder.JustApp):
...
+                "PYTHONUSERBASE":
+                "/env",  # this is adding the deps
+                "PATH":
+                "/env/bin:/opt/python%s/bin:/opt/python3.5/bin:" % py_version +
```
dlorenc: I think you might be able to use $PATH here instead of hardcoding the existing path elements.

ftl/python/builder.py
```diff
@@ -33,76 +33,106 @@ class Python(builder.JustApp):
...
+        logging.info("pkg_dirs" + str(pkg_dirs))
+        for pkg_dir in pkg_dirs:
+            layer, sha = self._gen_package_tar(pkg_dir)
+            logging.info('Generated layer with sha: %s', sha)
+            base_image = append.Layer(
+                base_image, layer, diff_id=sha, overrides=overrides)
+        return base_image
+
+    def _gen_dirs(self):
+        tmp_app = os.path.join(tempfile.mkdtemp(), 'app')
```
dlorenc: Could you make one tempdir then use it for all of these?

dlorenc: And maybe do this in a loop?

ftl/python/builder.py
```diff
@@ -33,76 +33,106 @@ class Python(builder.JustApp):
...
+        logging.info('Starting to gzip pip package tarfile...')
+        subprocess.check_call(['gzip', tar_path])
+        logging.info('Finished generating gzip pip package tarfile.')
+        return open(os.path.join(pkg_dir, tar_path + '.gz'), 'rb').read(), sha
 
+    def _gen_pip_env(self, venv_dir):
+        pip_env = os.environ.copy()
+        # bazel adds its own PYTHONPATH to the env
+        # which must be removed for the pip calls to work properly
+        pip_env.pop('PYTHONPATH', None)
```
dlorenc: You can use delete here instead of pop.

ftl/python/builder.py
aaron-prindle: if I just add the VIRTUAL_ENV var, python is not finding the packages installed when I attempt to run the python app in /app.  Adding PYTHONUSERBASE adds the correct deps and then the app runs.  Also not sure if it is acceptable but the app must be launched with python3.6 if using python3.6 even when the env vars are set.

ftl/python/builder.py
aaron-prindle: it wasn't clear to me as this is an override if this is called with context to what the PATH variable is currently

aaron-prindle:
all comments addressed

dlorenc:
Looks like the python test is failing: 
```shell
//ftl:python_builder_test                                                FAILED in 8.9s
```

ftl/python/builder.py
```diff
@@ -27,82 +27,109 @@
...
+        return base_image
+
+    def _gen_dirs(self, dirs):
+        tmp_dir = tempfile.mkdtemp()
+        dir_map = {}
+        for dir in dirs:
+            dir_name = os.path.join(tmp_dir, dir)
+            dir_map[dir] = dir_name
+            os.mkdir(dir_name)
+        return dir_map['app'], dir_map['env'], dir_map['wheel']
```
dlorenc: Return the whole dir_map? Or just the ones with entries in the "dirs" input?

ftl/python/builder.py
```diff
@@ -27,82 +27,109 @@
...
+        overrides = self._generate_overrides(py_version)
+
+        tmp_app, venv_dir, wheel_dir = self._gen_dirs(_WORK_DIRS)
+        whls = self._resolve_whls(tmp_app, venv_dir, wheel_dir)
+        pkg_dirs = [self._whl_to_fslayer(whl, venv_dir) for whl in whls]
+        logging.info("pkg_dirs" + str(pkg_dirs))
+        for pkg_dir in pkg_dirs:
+            layer, sha = self._gen_package_tar(pkg_dir)
+            logging.info('Generated layer with sha: %s', sha)
+            base_image = append.Layer(
```
dlorenc: I see that you're iteratively adding layers to "base_image" and re-using the same variable name, but it might be more clear if you make a copy once at the start:

```python
package_base = base_image
```

Then add and re-use package_base.

ftl/python/builder.py
```diff
@@ -27,82 +27,109 @@
...
+        return pip_env
+
+    def _resolve_whls(self,
+                      tmp_app,
+                      venv_dir,
+                      wheel_dir,
+                      py_version='python2.7'):
         # Copy out the relevant package descriptors to a tempdir.
         for f in self.descriptor_files:
             # if self._ctx.Contains(f):
```
dlorenc: Why is this commented out?

ftl/python/builder.py
dlorenc: How are you testing this? I think just the VIRTUAL_ENV part should work. Let's debug instead of using the other env var.

ftl/python/builder.py

ftl/python/builder.py

ftl/python/builder.py
aaron-prindle: I think this was a typo.  uncommented

aaron-prindle:
I have added a layer for the virtualenv base files and everything is working as expected now.

aaron-prindle:
kokoro passing, travis failed as 'wheel' isn't installed. adding dep to travis

ftl/python/builder.py
```diff
@@ -27,82 +28,132 @@
...
-    def CreatePackageBase(self, base_image, cache):
+    def CreatePackageBase(self, base_image, py_version='python2.7'):
         """Override."""
-        overrides = self._generate_overrides()
-
-        layer, sha = self._gen_package_tar()
-        logging.info('Generated layer with sha: %s', sha)
+        package_base = base_image
+
+        dir_map = self._gen_dirs(_WORK_DIRS)
```
dlorenc: Actually the map seems a little awkward here - what about just setting these on `self` and keeping the list of all dirs to create on self as well?

ftl/python/builder.py
```diff
@@ -27,82 +28,132 @@
...
+        self._setup_app_dir(dir_map['app'])
+        self._setup_venv(dir_map, py_version)
+        # copy /env to tmp_dir to get heirarchy correct
+        tmp_dir = tempfile.mkdtemp()
+        pkg_dir = os.path.join(tmp_dir, 'env')
+        # copy /env to pkg_dir
+        shutil.copytree(dir_map['env'], pkg_dir, True)
```
dlorenc: Why do we need this copy? Can we install into the pkg_dir directly?

ftl/python/builder.py
```diff
@@ -27,82 +28,132 @@
...
-        logging.info('Starting to gzip pip package tarfile...')
-        subprocess.check_call(['gzip', tar_path])
-        logging.info('Finished generating gzip pip package tarfile.')
-        return open(os.path.join(tmp_venv, tar_path + '.gz'), 'rb').read(), sha
+    def _whl_to_fslayer(self, whl, venv_dir):
+        tmp_dir = tempfile.mkdtemp()
+        pkg_dir = os.path.join(tmp_dir, 'env')
+        os.makedirs(pkg_dir)
+        subprocess.check_call(
+            ['pip', 'install', '--prefix', pkg_dir, whl],
```
dlorenc: Why do we need this second pip install? Can we re-use the one done during the initial pip wheel?

aaron-prindle: I believe the first pip install only fetches the wheel, this one is exploding them

ftl/python/builder.py
aaron-prindle: this was done to keep the tar generation the same as it removed root folder.  I have changed the implementation to support both keeping and removing root folder, the copy is gone now

ftl/python/builder.py

dlorenc:
Looks like a flake issue:

```
+ flake8 .
./ftl/python/builder.py:22:1: F401 'shutil' imported but unused

```

aaron-prindle:
kokoro passing, travis failing.  python_builder_test passing locally for me

aaron-prindle:
kokoro still hasn't run here

aaron-prindle:
it says that kokoro has passed in the build logs (click the details link).

PR #465, opened 2017-11-22
added support for local cloudbuild by making waitFor array

aaron-prindle:


aaron-prindle:
seems waitFor doesn't work properly in container-builder-local, was getting:
2017/11/22 15:01:00 Error loading config file: repeated string fields containing one element must be spe
cified like ['foo'], not 'foo'

dlorenc:
I think you might just have to put the 'waitFor': value in a list, something like: waitFor: ['x'] instead of waitFor: 'x'

aaron-prindle:
It's working with just changing waitFor to be an array, updated PR

aaron-prindle:
passing in Kokoro, Travis still hasn't started.  Merging

PR #467, opened 2017-11-29
Ftl php benchmark

priyawadhwa:


dlorenc:
```
Step #2 - "build-image-gcr.io/ftl-node-test/packages_test-image:latest":   Failed to clone https://github.com/silexphp/Pimple.git, git was not found, check that it is installed and in your PATH env.  
```

priyawadhwa:
Hmm so not sure why it failed initially, but I ran kokoro again and it passed

aaron-prindle:
Also we should probably make some benchmark libs as each runtime is going to have duped code.

priyawadhwa:
I got the small/medium/large apps by running the same queries on bigquery and taking the 20th, 50th and 95th percentiles in terms of number of required dependencies.

Definitely think a benchmark lib would be good!

PR #470, opened 2017-11-30
added python-version flag

aaron-prindle:


ftl/common/args.py
```diff
@@ -87,6 +88,20 @@ def extra_args(parser, opt_list):
                         final image'
             }
         ],
+        'python_version': [
+            '--python-version', {
+                "dest":
+                'python_vesrsion',
```
dlorenc: typo

ftl/common/args.py
```diff
@@ -87,6 +88,20 @@ def extra_args(parser, opt_list):
                         final image'
             }
         ],
+        'python_version': [
+            '--python-version', {
+                "dest":
```
dlorenc: Do the keys and values need to be on separate lines here? It would be easier to follow if they were inlined.

ftl/common/args.py
aaron-prindle: done

ftl/common/args.py
aaron-prindle: done

PR #472, opened 2017-12-01
Create benchmarking lib for FTL

priyawadhwa:
A benchmark library, since a lot of code was duplicated between Node and PHP


ftl/benchmark/args.py
```diff
@@ -0,0 +1,49 @@
...
+# limitations under the License.
+"""This package defines the shared cli args for ftl binaries."""
+
+import argparse
+
+
+def base_parser():
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument(
```
aaron-prindle: it might be worth using the same base parser as in common/args.py as most of the flags are the same.  Then you can make a sub parser here for just the benchmark related flags

ftl/php/benchmark/main.py
```diff
@@ -12,92 +12,32 @@
...
-import datetime
-import time
-import os
-import logging
-from google.cloud import bigquery
+from ftl.benchmark import args
+from ftl.benchmark import benchmark
+
 
 DATASET_NAME = 'ftl_benchmark'
```
aaron-prindle: The convention used in other code for private constants is to prefix with '_'.  It is the 'pythonic' way to say a variable is private.  These should probably be _DATASET_NAME, etc.

aaron-prindle:
Looks good! Left some comments.  Some unit tests for the shared lib might make sense here as well, right now no tests are verifying that the benchmarks are buildable/runnable or working as expected

ftl/benchmark/BUILD
```diff
@@ -0,0 +1,16 @@
+package(default_visibility = ["//visibility:public"])
+
+load(
+    "@io_bazel_rules_docker//python:image.bzl",
+    "py_image",
+)
+load("@base_images_docker//dockerfile_build:dockerfile_build.bzl", "dockerfile_build")
```
tejal29: not used in this build file.

ftl/benchmark/benchmark.py
```diff
@@ -0,0 +1,83 @@
...
+        logging.info('Beginning building {0} images'.format(self._runtime))
+        for _ in range(self._iterations):
+            start_time = time.time()
+
+            # For the binary
+            builder_path = 'ftl/{0}_builder.par'.format(self._runtime)
+
+            # For container builder
+            if not os.path.isfile(builder_path):
+                builder_path = 'bazel-bin/ftl/{0}_builder.par'.format(
```
tejal29: There is an underlying assumption here that .par file should be present. Its probably an output of the bazel rule.
There is a possibility even this file does not exists.
Can we put L61 to L83 inside a try Except OSError?
Also raise the exception with an appropriate message explaining the assumptions.

aaron-prindle: piggybacking on this, currently if the ftl.par file exists in the original repo, the build will fail.  a bazel clean is required to be run before launching the ftl_benchmark which shouldn't be the casel.  we should try to address that as well

ftl/benchmark/args.py
priyawadhwa: There are a few flags in common/args.py which aren't relevant to benchmarking, I feel like it might be more confusing to have extra flags that don't do anything?

ftl/php/benchmark/main.py
priyawadhwa: thanks!

ftl/benchmark/benchmark.py
priyawadhwa: yah, I'm not exactly sure how to because I don't think there's any way of only running the build step if there isn't a par file

ftl/benchmark/args.py
aaron-prindle: it's fine to keep as is.  The extra flags that don't do anything currently could be passed to the benchmark as well in the case that we at some point want to test them (as the benchmark is then just calling the builders with the flags passed).  This way if flags are added to the builders we only need to change them in once place.  Also if we wanted to test additional builder flags, they would already be passed to the benchmark.

priyawadhwa: Ahh that makes sense, I'll make the change then. Thanks!

aaron-prindle:
All of my PRs had the same issue for Travis, it seems that the Travis version of buildifier must have changed so it's identifying these files as incorrect.  #472 should fix that error

PR #474, opened 2017-12-02
refactor builders to share tar and gzip code

aaron-prindle:


dlorenc:
Some flake/formatting issues.

ftl/common/ftl_util.py
```diff
@@ -27,3 +31,18 @@ def __enter__(self):
     def __exit__(self, unused_type, unused_value, unused_traceback):
         end = time.time()
         logging.info('%s took %d seconds', self.descriptor, end - self.start)
+
+
+def folder_to_layer_sha(pkg_dir, pkg_mngr_name):
```
dlorenc: Maybe "zip_directory" would be more clear?

ftl/common/ftl_util.py
```diff
@@ -27,3 +31,18 @@ def __enter__(self):
...
+    tar_path = tempfile.mktemp()
+    with Timing("tar_%s_package" % pkg_mngr_name):
```
dlorenc: It's a bit weird we need the package manager name here - there's nothing else that uses this except the logging statement. Can we remove that as a parameter?

ftl/common/ftl_util.py
aaron-prindle: "zip_directory_to_layer_sha"?  This takes a path to a folder that is uncompressed/untarred

ftl/common/ftl_util.py
aaron-prindle: done

ftl/common/ftl_util.py
aaron-prindle: done

aaron-prindle:
also wanted to mention this is not every gzip call.  A similar thing is done in ftl/common/builder.py:
https://github.com/GoogleCloudPlatform/runtimes-common/blob/master/ftl/common/builder.py#L68-L87
It is different though in that it is retrieving files from a context that is not necessarily on disk.  For the actual FTL pipeline it might make sense to have this use the same method but we have MemoryContexts that wouldn't work a subprocess.check_call('tar', '...')

dlorenc:
> It is different though in that it is retrieving files from a context that is not necessarily on disk. For the actual FTL pipeline it might make sense to have this use the same method but we have MemoryContexts that wouldn't work a subprocess.check_call('tar', '...')

Ah, yeah. Maybe it would make sense to give the context interface a method to "tar/zip" itself up, similar to this. Then the disk context could take advantage of this optimization.

aaron-prindle:
I'll change this to be a part of the context interface

PR #477, opened 2017-12-08
Ftl benchmark python

priyawadhwa:


ftl/python/benchmark/main.py
```diff
@@ -0,0 +1,66 @@
...
+import argparse
+import sys
+from ftl.common import args
+from ftl.benchmark import benchmark
+
+
+_RUNTIME = "python"
+
+parser = args.base_parser()
+python_parser = argparse.ArgumentParser(
```
aaron-prindle: These benchmark flags are repeated in each benchmark main, it might be worth putting these in something like benchmark/args.py so they are not repeated for each language.

priyawadhwa: good idea, i made the change!

PR #478, opened 2017-12-11
[cloudbuild] Enable parallel builds

jprzychodzen:
* Add two flags, --enable_parallel and --force_parallel
* Change is backward compatibile

### DIFF
--- a/versioning/scripts/cloudbuild/main.go
+++ b/versioning/scripts/cloudbuild/main.go
@@ -61,7 +61,7 @@ const cloudBuildTemplateString = `steps:
       - 'build'
       - '--tag={{ .Tag }}'
       - '{{ .Directory }}'
-{{- if (eq $parallel true) }}
+{{- if $parallel }}
     waitFor: ['-']
     id: 'image-{{ .Tag }}'
 {{- end }}
@@ -99,7 +99,7 @@ const cloudBuildTemplateString = `steps:
       - 'UNIQUE={{ $imageIndex }}-{{ $testIndex }}'
       - '--test_spec'
       - '{{ $test }}'
-{{- if (eq $parallel true) }}
+{{- if $parallel }}
     waitFor: ['image-{{ $primary }}']
     id: 'test-{{ $primary }}-{{ $testIndex }}'
 {{- end }}
@@ -116,7 +116,7 @@ const cloudBuildTemplateString = `steps:
       - 'tag'
       - '{{ $primary }}'
       - '{{ . }}'
-{{- if (eq $parallel true) }}
+{{- if $parallel }}
     waitFor:
       - 'image-{{ $primary }}'
 {{- range $testIndex, $test := $image.FunctionalTests }}
@@ -136,7 +136,7 @@ images:
 timeout: {{ .TimeoutSeconds }}s
 {{- end }}
 
-{{- if (eq $parallel true) }}
+{{- if $parallel }}
 options:
   machineType: 'N1_HIGHCPU_32'
 {{- end }}
@@ -165,10 +165,8 @@ type cloudBuildTemplateData struct {
 	TimeoutSeconds int
 }
 
-func parallelize(
-	options cloudBuildOptions,
-	numberOfVersions int,
-	numberOfTests int) bool {
+func shouldParallelize(
+	options cloudBuildOptions, numberOfVersions int, numberOfTests int) bool {
 	if options.ForceParallel {
 		return true
 	}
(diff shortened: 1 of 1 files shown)


versioning/scripts/cloudbuild/main.go
```diff
@@ -251,6 +296,8 @@ func main() {
 	newTagsPtr := flag.Bool("new_tags", false, "Require that image tags do not already exist.")
 	firstTagOnly := flag.Bool("first_tag", false, "Build only the first per version.")
 	timeoutPtr := flag.Int("timeout", 0, "Timeout in seconds. If not set, the default Cloud Build timeout is used.")
+	enableParallel := flag.Bool("enable_parallel", false, "Enable parallel build and bigger VM")
```
huyhg: shouldn't this be `true` by default? we still respect what's in the config file right, which by default is off

versioning/scripts/cloudbuild/main.go
```diff
@@ -129,11 +159,25 @@ type imageBuildTemplateData struct {
 
 type cloudBuildTemplateData struct {
 	RequireNewTags bool
+	Parallel       bool
 	ImageBuilds    []imageBuildTemplateData
 	AllImages      []string
 	TimeoutSeconds int
 }
 
+func parallelize(
```
huyhg: rename to `shouldParallelize`

versioning/scripts/cloudbuild/main.go
```diff
@@ -129,11 +159,25 @@ type imageBuildTemplateData struct {
...
+	options cloudBuildOptions,
```
huyhg: all args should just be on one line

versioning/scripts/cloudbuild/main.go
```diff
@@ -54,6 +61,10 @@ const cloudBuildTemplateString = `steps:
       - 'build'
       - '--tag={{ .Tag }}'
       - '{{ .Directory }}'
+{{- if (eq $parallel true) }}
```
huyhg: should these just be `if $parallel`?

jprzychodzen: Done.

I was trying to do something like that at beginning, but it failed. Now I cannot check difference.

versioning/scripts/cloudbuild/main.go
jprzychodzen: Done.

versioning/scripts/cloudbuild/main.go
jprzychodzen: Done.

versioning/scripts/cloudbuild/main.go
jprzychodzen: I do want to keep this change as optional, with backward compatibility. It requires tests to parallelable, so this change can possibly break builds that depend on sequential execution.

huyhg: I understand this as: even when this flag is `true`, unless we change the config files, no parallelization will automatically happen. `forceParallel` on the other hand will break non-ready tests.

versioning/scripts/cloudbuild/main.go
```diff
@@ -165,10 +165,8 @@ type cloudBuildTemplateData struct {
 	TimeoutSeconds int
 }
 
-func parallelize(
-	options cloudBuildOptions,
-	numberOfVersions int,
-	numberOfTests int) bool {
+func shouldParallelize(
+	options cloudBuildOptions, numberOfVersions int, numberOfTests int) bool {
```
huyhg: to previous line as well

versioning/scripts/cloudbuild/main.go
jprzychodzen: What kind of config file you have in mind? I do not use any. Please tell me more about this feature.

In this patch logic is as follows:
- no flag - as is, serial build,
- enable flag - if 'big repo', then parallel build, else serial build,
- force flag or both flags - parallel build.

Changing enable_parallel to default true:
- no flag - if 'big repo', then parallel build, else serial build,
- negate enable flag - as is, serial build,
- force flag - parallel build.

versioning/scripts/cloudbuild/main.go
jprzychodzen: Done.

versioning/scripts/cloudbuild/main.go
huyhg: My bad--for whatever reason I thought our `versions.yaml` has a bit indicating whether it desires to be parallelized. I see that 'big repo' is defined in `shouldParallelize`.

PR #479, opened 2017-12-12
update ftl_release to upload flt.par to :latest

aaron-prindle:


ftl/ftl_release.yaml
```diff
@@ -8,6 +8,9 @@ steps:
     args: ['build','//ftl:node_builder.par']
   - name: 'gcr.io/cloud-builders/gsutil'
     args: ['cp', 'bazel-bin/ftl/node_builder.par', 'gs://gcp-container-tools/ftl/node/${COMMIT_SHA}/ftl.par']
+  - name: 'gcr.io/cloud-builders/gsutil'
```
dlorenc: Tiny nit: If you copy from the cloud location you just copied to above, it'll be much faster.

aaron-prindle: done

PR #483, opened 2017-12-21
change /srv to be default directory for all app files

aaron-prindle:


ftl/common/builder.py
```diff
@@ -69,7 +69,7 @@ def BuildAppLayer(self):
         with tarfile.open(fileobj=buf, mode='w') as out:
             for name in self._ctx.ListFiles():
                 content = self._ctx.GetFile(name)
-                info = tarfile.TarInfo(os.path.join('app', name))
+                info = tarfile.TarInfo(os.path.join('srv', name))
```
squee1945: Why is `srv` hard-coded? Shouldn't it come from the flag value?

aaron-prindle: right now value for '--destination' is passed to only one step of the builder and BuildAppLayer does not have access to the args.  I just updated the PR to have BuildAppLayer accept the destination_path as a quick fix. Do you require app directory to be configurable with a flag or is simply changing the default to /srv is enough for your use case?

PR #486, opened 2018-01-08
Php phs 2

aaron-prindle:


aaron-prindle:
when you composer require pkg version composer installs multiple files and updates the composer.json. it might be worth deleting all of the generated files except for what is is in /vendor

aaron-prindle:
PTAL

ftl/php/builder.py
```diff
@@ -33,27 +34,38 @@ def __init__(self, ctx, args, cache_version_str):
         super(PHP, self).__init__(ctx, _PHP_NAMESPACE, args, cache_version_str,
                                   [_COMPOSER_LOCK, _COMPOSER_JSON])
 
+    def _parse_composer_pkgs(self):
+        descriptor_contents = ftl_util.descriptor_parser(
+            self._descriptor_files, self._ctx)
+        composer_json = json.loads(descriptor_contents)
+        pkgs = []
+        for k, v in composer_json['require'].iteritems():
+            pkgs.append(k + ' ' + v)
```
dlorenc: This feels a bit hacky - could we keep this as a tuple or some other structure? It looks like you split this back apart by space later on anyway.

aaron-prindle:
getting timings locally:
full build of this composer.json:
https://gist.github.com/aaron-prindle/588998e1fe679af669cf5e92027b9a52

`2018-01-17,14:27:45.102 INFO     full build took 50 seconds`

rebuild with updated composer.json w/ single package:
https://gist.github.com/aaron-prindle/ef8ac052640eddca26fd1fc725b5734d

`2018-01-17,14:31:01.865 INFO     full build took 23 seconds`

ftl/php/builder.py
```diff
@@ -33,27 +34,38 @@ def __init__(self, ctx, args, cache_version_str):
...
+            for pkg_txt in pkgs:
+                pkg = self.PackageLayer(self._ctx, self._descriptor_files,
+                                        pkg_txt, self._args.destination_path)
+                cached_pkg_img = None
                 if self._args.cache:
-                    with ftl_util.Timing("uploading pkg layer"):
-                        self._cash.Store(self._base, self._namespace,
-                                         pkg.GetCacheKey(), pkg.GetImage())
-            lyr_imgs.append(pkg)
+                    with ftl_util.Timing("checking cached pkg layer"):
```
dlorenc: nit: i think you used underscores in here in other places?

ftl/php/builder.py
```diff
@@ -93,8 +105,9 @@ def _gen_composer_install_tar(self, pkg_descriptor, destination_path):
             os.makedirs(app_dir)
 
             # Copy out the relevant package descriptors to a tempdir.
-            ftl_util.descriptor_copy(self._ctx, self._descriptor_files,
-                                     app_dir)
+            if pkg_descriptor is None:
```
dlorenc: not sure I understand this part. Why do we only copy if it's none?

aaron-prindle: if there is no pkg_descriptor, that means that composer is being run without a pkg in mind but with a composer.json file in its entirety.  When using the composer.json file, we need to have it copied.  When using a pkg_descriptor, we don't need to copy but can just use the string arg:
composer require <pkg> <version>
VS
composer install   # this uses composer.json in same dir

ftl/php/builder.py
aaron-prindle: i think I may have undid this in all places, so this is consistent currently.  Should I change all of these prints to be underscored?

dlorenc: nah, it's fine as long as it's all consistent.

ftl/php/builder.py
aaron-prindle: I changed this to be a tuple

aaron-prindle:
yes, the tests ensure that packages are still being installed correctly, the flags are still working properly, and images are still being uploaded.  The caching test was done separately to verify that this method works as expected, I can verify that only uncached packages are built on reupload

PR #487, opened 2018-01-10
support multiple layers for FromFSImage

aaron-prindle:


aaron-prindle:
@sharifelgamal the tar_to_dockerimage.py, FromFSImage class used as a standalone tool should be what you need after this PR

PR #490, opened 2018-01-18
Add the cred helper layers to the two new base images.

mattmoor:


mattmoor:
@dlorenc FYI

dlorenc:
@mattmoor Just FYI, these images aren't actually used by anyone other than our own tests, AFAIK.

@sqee1945's team is producing their own base images that act as builders, including the FTL code as a .par. 

I don't mind adding the cred helpers, but before you take a real dependency on these we should have a chat :)

mattmoor:
@dlorenc I know, I wanted something to test and this was quicker to fix. :)

PR #491, opened 2018-01-22
entrypoint flag added to node

aaron-prindle:


aaron-prindle:
PTAL, entrypoint test passing

aaron-prindle:
talked offline, should add entrypoint flag functionality to python and php runtimes as well.  Also add tests for those runtimes.

squee1945:
I like that; this flag has general usefulness.

PR #492, opened 2018-01-23
Add extensions and configuration to dockerfiles

jprzychodzen:


jprzychodzen:
Sample usage of these extensions https://github.com/GoogleCloudPlatform/postgresql-docker/commit/2fd6772b6ef97d1e664e4df20ddf4253624b6780

versioning/scripts/dockerfiles/main.go
```diff
@@ -15,14 +15,55 @@ import (
 	"text/template"
 
 	"github.com/GoogleCloudPlatform/runtimes-common/versioning/versions"
+	"strconv"
```
huyhg: wrong order of imports. please ensure gofmt is run

versioning/scripts/dockerfiles/main.go
```diff
@@ -15,14 +15,55 @@ import (
...
 )
 
+type Format struct {
```
huyhg: Can we add some doc to this struct? It's not obvious what `Count` is for

versioning/scripts/dockerfiles/main.go
```diff
@@ -15,14 +15,55 @@ import (
...
+	Indent string
+	Count  int
+}
+
+func parseSpecToFormat(c map[string]string) Format {
+	fIndent := " "
+	fCount := 4
+	if val, ok := c["template_retry_intendt_type"]; ok {
```
huyhg: what is `intendt_type`?

versioning/scripts/dockerfiles/main.go
```diff
@@ -15,14 +15,55 @@ import (
...
+	return str
+}
+
 func renderDockerfile(version versions.Version, tmpl template.Template) []byte {
 	var result bytes.Buffer
 	tmpl.Execute(&result, version)
 	return result.Bytes()
 }
 
+func renderRetry(pack versions.Package, tmpl template.Template) string {
```
huyhg: s/pack/pkg or s/pack/package

versioning/versions/versions.go
```diff
@@ -63,3 +64,12 @@ func validateUniqueTags(spec Spec) {
 		}
 	}
 }
+
+func (spec *Spec) CheckExtension(s string) bool {
```
huyhg: `HasExtension`

versioning/versions/versions.go
```diff
@@ -31,7 +32,9 @@ type Version struct {
 }
 
 type Spec struct {
-	Versions []Version
+	Versions   []Version
+	Extensions []string
```
huyhg: instead of a list of strings, I rather have this explicitly as fields in the spec, so we can write something like this:
```
spec:
  versions:
    [...]
  extensions:
    from: {}  # from has no configurations
    retry:
      indent: '    '
```

Having config, which is detached from the extensions, is undesirable.

versioning/scripts/dockerfiles/main.go
jprzychodzen: gofmt had no problems, manually moved into top imports

versioning/scripts/dockerfiles/main.go
jprzychodzen: Removed count, merging it's function with 'indent_type' as 'indent'.

versioning/scripts/dockerfiles/main.go
jprzychodzen: Merged with count, now this is a user-defined intendt.

versioning/scripts/dockerfiles/main.go
jprzychodzen: s/pack/package done

versioning/versions/versions.go
jprzychodzen: Done

versioning/versions/versions.go
jprzychodzen: Done

versioning/scripts/dockerfiles/main.go
```diff
@@ -170,6 +203,25 @@ func check(e error) {
 	}
 }
 
+const templateFrom = `FROM {{ .From }}
+`
+
+const templateRetry = `found='' && \
+ for server in \
```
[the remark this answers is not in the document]
huyhg: I see what this does now. However, this is assuming that we always want 1 indent character in the result, which is untrue.

versioning/versions/versions.go
```diff
@@ -31,7 +32,8 @@ type Version struct {
 }
 
 type Spec struct {
-	Versions []Version
+	Versions   []Version
+	Extensions map[string]map[string]string
```
huyhg: OK, I still feel the extension mechanism here unsatisfying. It's hard to connect what defined the YAML file and what the template does.

I have a proposal for this here: https://github.com/GoogleCloudPlatform/runtimes-common/commit/f6ed5287162a8c42d4c2212e51fc179e428043cf

I think the `FROM` feature is simple enough for us to not worry about. My proposal is essentially introducing a function that the template can freely use instead. The template change can be found here: https://github.com/GoogleCloudPlatform/postgresql-docker/commit/f75608afada442cf848ee2ef1a1a83e997a9ed6c

jprzychodzen:
Dropped in favor of https://github.com/GoogleCloudPlatform/runtimes-common/pull/549

PR #507, opened 2018-02-13
Switch to dep instead of GoDeps as GoDeps is now archived

tejal29:
I was trying to add the cobra dependency using GoDep and realized this project is now archived.

Switching this over to use dep instead.


Gopkg.toml
```diff
@@ -0,0 +1,38 @@
...
+  name = "github.com/ghodss/yaml"
+  version = "1.0.0"
+
+[[constraint]]
+  name = "github.com/golang/protobuf"
+  version = "1.0.0"
+
+[prune]
+  go-tests = true
+  unused-packages = true
```
r2d4: I also use ` go-tests = true` to prune vendored tests

PR #511, opened 2018-02-20
added cached tests

aaron-prindle:


aaron-prindle:
cache test uploads same image twice, testing that layers are the same with the exception of a single app layer which differs on each upload (all dep layers should be the same).  Currently the cloudbuild step is added as a separate run, similar to the benchmark test and not included as part of the integration tests but this would be easy to change.

ftl/node/cached/main.py
```diff
@@ -0,0 +1,35 @@
+# Copyright 2017 Google Inc. All rights reserved.
```
tejal29: 2018

ftl/cached/cached_test.py
```diff
@@ -0,0 +1,43 @@
+# Copyright 2017 Google Inc. All Rights Reserved.
```
tejal29: 2018

ftl/cached/args.py
```diff
@@ -0,0 +1,53 @@
+# Copyright 2017 Google Inc. All Rights Reserved.
```
tejal29: 2018

ftl/cached/cached.py
```diff
@@ -0,0 +1,122 @@
+# Copyright 2017 Google Inc. All rights reserved.
```
tejal29: 2018

ftl/cached/ftl_node_cached_yaml.py
```diff
@@ -0,0 +1,72 @@
...
+                ]
+            },
+            # Get image digests so that we can delete uploaded images
+            {
+                'name':
+                'gcr.io/cloud-builders/gcloud',
+                'entrypoint':
+                '/bin/bash',
+                'args': [
+                    '-c', 'gcloud container images list-tags ' + path + ' | ' +
```
dlorenc: Can you do this part in the test itself?

ftl/node/cached/BUILD
```diff
@@ -0,0 +1,42 @@
...
+    main = "main.py",
+    deps = [
+        "//ftl:ftl_lib",
+        "//ftl/cached:cached_lib",
+    ],
+)
+
+load("@base_images_docker//dockerfile_build:dockerfile_build.bzl", "dockerfile_build")
+
+dockerfile_build(
```
dlorenc: Why do we need a dockerfile build here?

ftl/cached/cached.py
```diff
@@ -0,0 +1,105 @@
...
+                logging.info('output of build {0}: {1}'.format(label, output))
+                lyr_shas.append(self._fetch_lyr_shas(img_name))
+                self._del_img_from_gcr(img_name)
+            try:
+                self._compare_layers(lyr_shas[0], lyr_shas[1])
+            except RuntimeError as e:
+                logging.error(e)
+                exit(1)
+        except OSError:
+            raise OSError("""Cached test assumes either ftl/{0}_builder.par
```
dlorenc: This shouldn't be an oserror. You're also swallowing the real error here, which might be anything.

ftl/cached/ftl_python_cached_yaml.py
```diff
@@ -0,0 +1,72 @@
+"""A script to generate a cloudbuild yaml."""
+
+import os
+import yaml
+
+_TEST_DIR = '/workspace/ftl/python/testdata'
```
dlorenc: Could you refactor the three new cache .py files into one script that takes the language as a parameter?

ftl/node/cached/BUILD
aaron-prindle: removed

ftl/cached/cached.py
aaron-prindle: removed

ftl/cached/ftl_python_cached_yaml.py
aaron-prindle: done

ftl/cached/ftl_node_cached_yaml.py
aaron-prindle: done

ftl/cached/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM bazel/reproducible:ubuntu_vanilla
```
dlorenc: Why do we need this Dockerfile?

aaron-prindle: we don't, i will remove it in another PR as kokoro is a bit flaky rn

PR #512, opened 2018-02-20
removing entrypoint detection/setting in node builder

aaron-prindle:


aaron-prindle:
with this change, images are confirmed working on GCF

tejal29:
Is there a reason why we are removing the entry point detection?

aaron-prindle:
talked offline.  this is being removed as some consumer of ftl have an entrypoint set in the base image that would be overwritten everytime by this entrypoint detection piece.  we still have an --entrypoint flag that can be set to allow overwriting the final entrypoint of an FTL image.  This is to allow for the use case of setting an entrypoint in the base image and having it propogated.

PR #513, opened 2018-02-21
add destination path to cache key

aaron-prindle:


aaron-prindle:
fixes the issue where a destination path can be specified and cached but then subsequent images with different destination paths will fail using cached deps as they will be in the wrong place

PR #517, opened 2018-02-22
Fix error where .npmrc is not being added before running npm install

### LINKED ISSUE #516: Add support for private packages with node
As discussed on the [GAE Standard Node.js EAP group ](https://groups.google.com/d/msg/appengine-standard-nodejs-eap/SlDXBm3uRgs/gzl4BCkAAQAJ)(requires access), there is an issue with `.npmrc` not being copied over by ftl before an image is built. Issue reported by Troy Kelly. This breaks installing private packages from npm.

One fix would be to add `.npmrc` to the list of files being [initially copied](https://github.com/GoogleCloudPlatform/runtimes-common/blob/a96c4e0d4c8ea658ae0c93180d3f38dea35ef7d6/ftl/node/builder.py#L28) by ftl during a build. There might be security concerns with this. For example due to `unsafe-perm`. I have no idea about all of the use cases of ftl internally at Google.

I'd love to work on this. Is it safe to do what I outlined above?

hvaara:
Fixes #516


hvaara:
Added a PR. I am going to write tests for this as well.

dlorenc:
Awesome. Thanks for the PR! Let me know once you have the tests written and we can get this merged.

hvaara:
@dlorenc I implemented it as an integration test and tested it on my own project. Build/tests passed. Probably not the most elegant, but I wasn't sure how to do it otherwise. Not very familiar with FTL.

PR #522, opened 2018-02-23
Display pip stdout and stderr when installing wheels, and check return code

myelin:
- Always log stdout
- Log stderr if it's non-empty
- Raise an exception if the `pip wheel` command exits with a nonzero return code


aaron-prindle:
awesome, thanks for this!  I'm going to use the same logic and wrap it around all of the runtimes subprocess calls.

PR #537, opened 2018-03-02
fix caching args to be evaluated in cache obj

aaron-prindle:


ftl/common/cache.py
```diff
@@ -107,6 +111,8 @@ def _tag(self, cache_key, repo=None):
             tag=hashlib.sha256(fingerprint).hexdigest()))
 
     def Get(self, cache_key):
+        if not self._should_cache:
```
dlorenc: Can you add a logging statement in both of these if statements?

aaron-prindle: done

PR #541, opened 2018-03-05
make ftl support implicit :latest tag for --name

aaron-prindle:


ftl/common/builder.py
```diff
@@ -82,7 +82,7 @@ def __init__(self, ctx, namespace, args, cache_version_str,
         self._base_name = docker_name.Tag(self._args.base)
```
dlorenc: Could you update this one as well?

aaron-prindle: done

aaron-prindle:
fixed with #541

PR #545, opened 2018-03-06
add manual ftl versioned release cloudbuild job

aaron-prindle:


aaron-prindle:
yea I think it makes more sense to launch this job with a cloudbuild trigger so we have a branch for each version.  I will split it separately though into 3 different .yamls for each runtime though so that the cloudbuild tag regex can version each runtime seperately

aaron-prindle:
added 3 cloudbuild.yaml files that trigger a build based on the tag name pushed.  the regex for the git tag's are:
node-v[\d].[\d].[\d] -> node release
php-v[\d].[\d].[\d] -> php release
python-v[\d].[\d].[\d] -> python release

PR #548, opened 2018-03-07
support for many custom commands addded to dockerfiles

rafalbiegacz:
This functionality will allow to add many custom actions/commands to Dockerfile.templates. Once approved it will obsolute 'Cmd' field from versions.yaml which supports only one command per a given version.

### DIFF
--- a/versioning/versions/versions.go
+++ b/versioning/versions/versions.go
@@ -20,25 +20,13 @@ type Package struct {
 	Md5     string
 }
 
-// this will allow for passing many custom commands to Dockerfile.template
-// usage in versions.yaml
-//    custom:
-//       your_command:
-//                  cmd: "<command to be added to Dockerfile>"
-// usage in Dockerfile.template
-// {{- $cmd1 := index .Custom "your_command" }}
-// {{ $cmd1.Cmd }} \
-type Command struct {
-	Cmd string
-}
-
 type Version struct {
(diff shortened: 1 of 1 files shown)


versioning/versions/versions.go
```diff
@@ -26,6 +26,7 @@ type Version struct {
 	Tags         []string
 	From         string
 	Cmd          string
+	TemplateArgs map[string]string
```
huyhg: We can call this `Values` to avoid redundant "template" when this is used from the template. Essentially, we'll have

```
{{- $custom := index .Values "debian_repo" }}
```

instead of

```
{{- $custom := index .TemplateArgs "debian_repo" }}
```

versioning/versions/versions.go
```diff
@@ -26,6 +26,7 @@ type Version struct {
 	Tags         []string
 	From         string
 	Cmd          string
```
huyhg: Follow up to remove `Cmd`?

versioning/versions/versions.go
huyhg: Or if you want to keep `TemplateArgs`, please add `` `yaml:"templateArgs"` `` annotation (similar to `ExcludeTests`) so that in the yaml file, we can use `templateArgs` instead of `templateargs`.

PR #550, opened 2018-03-08
Support for subfolders in templates directory

rafalbiegacz:
if one adds the following tag to a given version in versions.yaml:
    "templatedir: templates_subfolder"
then Dockerfile.template will be taken from <templates_dir>/templates_subfolder instead of a Dockerfile.template from <templates_dir>


huyhg:
Hey, I haven't been added as a reviewer yet. Let me know if it's ready for reviewing by adding me as the reviewer. Note that the github UI is kind of confusing: my name showing up but under "suggestions" looks like I've been added, but that's just a suggestion :-/

versioning/versions/versions.go
```diff
@@ -22,6 +22,7 @@ type Package struct {
 
 type Version struct {
 	Dir          string
+	TemplateDir  string
```
huyhg: There are 2 thoughts:
- We can't put shared files in the root `template/` folder. This could be useful for some files that are version agnostic.
- We could make this an array to pick and choose which sub-dir to include.

Both of these can be potentially useful but I'm OK with implementing them in the future.

versioning/scripts/dockerfiles/main.go
```diff
@@ -173,30 +180,39 @@ func check(e error) {
 func main() {
 	templateDirPtr := flag.String("template_dir", "templates", "Path to directory containing Dockerfile.template and any other files to copy over")
 	verifyPtr := flag.Bool("verify_only", false, "Verify dockerfiles")
+	var templatePath string
```
huyhg: not sure what this block is doing?

versioning/scripts/dockerfiles/main.go
```diff
@@ -173,30 +180,39 @@ func check(e error) {
...
-		for _, version := range spec.Versions {
+		if *verifyPtr {
+			failureCount := verifyDockerfiles(spec, *tmpl)
+			failureCount += verifyCopiedFiles(spec, *templateDirPtr)
+			os.Exit(failureCount)
+		} else {
 			data := renderDockerfile(version, *tmpl)
 			writeDockerfile(version, data)
-			copyFiles(version, *templateDirPtr)
+			copyFiles(version, templatePath)
```
huyhg: `templateDir`

versioning/scripts/dockerfiles/main.go
```diff
@@ -173,30 +180,39 @@ func check(e error) {
...
 
 	var spec versions.Spec
 	spec = versions.LoadVersions("versions.yaml")
 
-	templatePath := filepath.Join(*templateDirPtr, "Dockerfile.template")
-	templateData, err := ioutil.ReadFile(templatePath)
-	templateString := string(templateData)
-	check(err)
+	for _, version := range spec.Versions {
+		if version.TemplateDir != "" {
```
huyhg: I think we're gonna need an extra variable `templateDir` which would evaluate to either
- `filepath.Join(*templateDirPtr, version.TemplateDir)`, or
- `*templateDirPtr`

And `templatePath` should simply be `filepath.Join(templateDir, "Dockerfile.template")`.

versioning/versions/versions.go
rafalbiegacz: We still can pick files from: templates folder (e.g. files to be shared between different versions) and a specific subfolder. It should be enough.

If we go with a array/list of subfolders then we will need to implement add'l logic to select specific files from a given subfolder. This will add add'l complexity - not sure if we need it.

versioning/scripts/dockerfiles/main.go
rafalbiegacz: templatePath is to represent a path to Dockerfile.template within a specific subfolder.
templatePath is different from templateDirPtr - the latter points to 'templates' folder, the former points to a subfolder of 'templates'

versioning/scripts/dockerfiles/main.go
rafalbiegacz: Here we should use the variable that points to subfolder storing Dockerfile.template - in this case this variable is templatePath.

huyhg: If I read the code correctly, `templatePath` is actually the path to the Dockerfile template. This is why in my other comment I suggested adding another variable `templateDir` for this method.

rafalbiegacz: Good catch! I corrected the issue.

versioning/scripts/dockerfiles/main.go
```diff
@@ -227,33 +230,41 @@ func check(e error) {
...
 
-	if *verifyPtr {
-		failureCount := verifyDockerfiles(spec, *tmpl)
-		failureCount += verifyCopiedFiles(spec, *templateDirPtr)
-		os.Exit(failureCount)
-	} else {
-		for _, version := range spec.Versions {
+		if *verifyPtr {
+			failureCount := verifyDockerfiles(spec, *tmpl)
+			failureCount += verifyCopiedFiles(spec, *defaultTemplateDirPtr)
```
huyhg: This construct is strange, as `verifyCopiedFiles` loops through the versions again, and this block ends with an `os.Exit()`. Not related to this PR--would be nice to refactor this into something more sensible.

versioning/scripts/dockerfiles/main.go
```diff
@@ -227,33 +230,41 @@ func check(e error) {
 }
 
 func main() {
-	templateDirPtr := flag.String("template_dir", "templates", "Path to directory containing Dockerfile.template and any other files to copy over")
+	defaultTemplateDirPtr := flag.String("template_dir", "templates", "Path to directory containing Dockerfile.template and any other files to copy over")
 	verifyPtr := flag.Bool("verify_only", false, "Verify dockerfiles")
+	var version = "0.01"
```
huyhg: I kind of understand what we want to achieve here. This however doesn't seem very useful as no one is gonna read this output. And it's not related to the change this PR is trying to make.

My suggestion is dropping this from this PR. We can open a separate PR that solves this concern better, for example `versions.yaml` can specify a required sem ver of the app.

rafalbiegacz: agreed; will remove it.

versioning/scripts/dockerfiles/main.go
```diff
@@ -227,33 +225,36 @@ func check(e error) {
 }
 
 func main() {
-	templateDirPtr := flag.String("template_dir", "templates", "Path to directory containing Dockerfile.template and any other files to copy over")
+	defaultTemplateDirPtr := flag.String("template_dir", "templates", "Path to directory containing Dockerfile.template and any other files to copy over")
 	verifyPtr := flag.Bool("verify_only", false, "Verify dockerfiles")
+	var failureCount int
```
huyhg: `failureCount := 0`

versioning/scripts/dockerfiles/main.go
```diff
@@ -227,33 +225,36 @@ func check(e error) {
...
+		check(err)
 
-	if *verifyPtr {
-		failureCount := verifyDockerfiles(spec, *tmpl)
-		failureCount += verifyCopiedFiles(spec, *templateDirPtr)
-		os.Exit(failureCount)
-	} else {
-		for _, version := range spec.Versions {
+		if *verifyPtr {
+			failureCount = verifyDockerfiles(version, filepath.Join(*defaultTemplateDirPtr, version.TemplateSubDir), *tmpl)
```
huyhg: `failureCount += verifyDockerfiles`

PR #551, opened 2018-03-08
Add Bazel support for ctc lib and Hook it up CI

tejal29:
This PR is based on  https://github.com/GoogleCloudPlatform/runtimes-common/pull/533

Please review https://github.com/GoogleCloudPlatform/runtimes-common/pull/551/commits/3d5dd0a8e1d68b6d0673d1646db7cb9bf367e4e5

PR #556, opened 2018-03-09
GO CLI Lib Add default Templates. Refactor Execute and define Interface Verified

tejal29:
Depends on PR  #551  and #533 

Review Commit 
https://github.com/GoogleCloudPlatform/runtimes-common/pull/556/commits/814fea1f4d5cc3f25eccd2b9e399591a5f1850a1

### DIFF
--- a/ctc_lib/CLI_interface.go
+++ b/ctc_lib/CLI_interface.go
@@ -1,0 +1,66 @@
+/*
+Copyright 2018 Google Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package ctc_lib
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+type CLIInterface interface {
+	PrintO(c *cobra.Command, args []string)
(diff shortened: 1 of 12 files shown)


ctc_lib/CLI_interface.go
```diff
@@ -0,0 +1,66 @@
+/*
+Copyright 2018 Google Inc. All Rights Reserved.
```
dlorenc: Could you name this all lowercase?

dlorenc: The file, not the structs.

ctc_lib/CLI_interface.go
```diff
@@ -0,0 +1,66 @@
...
+	ValidateCommand() error
+	IsRunODefined() bool
+	Init()
+}
+
+func Execute(ctb CLIInterface) (err error) {
+	defer errRecover(&err)
+
+	if err := ctb.ValidateCommand(); err != nil {
+		CommandExit(err)
```
dlorenc: This function can return an error, should we do that here instead of exiting?

ctc_lib/CLI_interface.go
```diff
@@ -0,0 +1,66 @@
...
+			ctb.PrintO(c, args)
+		}
+		ctb.SetRun(cobraRun)
+	}
+
+	err = ctb.GetCommand().Execute()
+
+	//Add empty line as template.Execute does not print an empty line
+	ctb.GetCommand().Println()
+	CommandExit(err)
```
dlorenc: It's a little weird to collect the error, then call exit, then return it. Can CommandExit return the error of it doesn't exit for some reason? Then we can return CommandExit

ctc_lib/CLI_interface.go
tejal29: done

ctc_lib/CLI_interface.go
tejal29: yeah, i think i wanted to define ExecuteE which returns Error and Execute() which does os.Exit.
  Forgot to make those changes.

tejal29: done

PR #560, opened 2018-03-15
Add Logging Support for CLI Lib.

tejal29:
The Logging Module I have used is Logrus.
I read an article which listed pros and cons of some of the GO Log modules used.
Logrus has a lot of rich features like you can Set custom formatters,  redirect output to elsewhere, Add hooks.
Glog was another potential candidate. But i chose Logrus over this.

In this Pr, 
1. Initialize logging
2. Set the log level from Command line
3. Add Hooks for Log.Panic and Log.Fatal to use Command.Exit.

TODO:
Add logging DIR to write logs in a Dir. Probably add this in with the Viper Configs PR coming up.


ctc_lib/container_tool_command_base.go
```diff
@@ -67,3 +76,32 @@ func (ctb *ContainerToolCommandBase) ReadTemplateFromFlagOrCmdDefault() string {
 	}
 	return flags.TemplateString
 }
+
+// The command line flags are not parsed until Execute is called.
+// Hence all logic to init global flags should happen in PersistentPreRun.
+func (ctb *ContainerToolCommandBase) initGlobalFlags() {
+	init_persisted_flags := func(cmd *cobra.Command, args []string) {
```
dlorenc: nit: camelCase

ctc_lib/container_tool_command_base.go
```diff
@@ -67,3 +76,32 @@ func (ctb *ContainerToolCommandBase) ReadTemplateFromFlagOrCmdDefault() string {
...
+		Log.SetLevel(flags.LogLevel.Level)
+	}
+	// Set PersistentPreRun to above function if its not set already.
+	// Else overwrite the existing one with init_persisted_flags followed by
+	// user set.
+	if ctb.PersistentPreRun == nil && ctb.PersistentPreRunE == nil {
```
dlorenc: It might be cool to support existing PreRuns as well. You could do something like:

```go
originalPrerun := ctb.PersistentPrerun

ctb.PersistentPrerun = func(cmd *cobra.Command, args []string) {
  originalPrerun()
  initPersistentFlags()
}
```

tejal29: Actually i just found out, i dont need this any more. 
I registered a function using cobra.OnInitialize() which is called after Command.ParseFlags().
its changed in https://github.com/GoogleCloudPlatform/runtimes-common/pull/564. Will make that change here too and update the PR.

PR #566, opened 2018-03-21
Add a command to print all command configurations and Set config through command line.

tejal29:
This is based of PR https://github.com/GoogleCloudPlatform/runtimes-common/pull/565

Please review https://github.com/GoogleCloudPlatform/runtimes-common/commit/5bc4504ab5c0a2c4666112adc4652ff1daa04648 commit 

This command is used in update check. Breaking this into a smaller change here.


tejal29:
I just realized WriteConfig in set is a bad idea since it will overwrites any comments and notes

ctc_lib/config_command_test.go
```diff
@@ -0,0 +1,101 @@
...
+			Phase: "test",
+		},
+		Output: nil,
+		RunO: func(command *cobra.Command, args []string) (interface{}, error) {
+			return nil, nil
+		},
+	}
+
+	Version = "1.0.1"
+	ConfigFile = "testdata/testConfig.json"
```
aaron-prindle: is this used in the test?

tejal29: These are Package Variables that Container Command Tool Lib Uses. 
These Package Variables are defined at https://github.com/GoogleCloudPlatform/runtimes-common/blob/master/ctc_lib/ctc_lib.go#L33

An Example setting these values is in Demos.
https://github.com/GoogleCloudPlatform/runtimes-common/blob/master/demo/ctc/cmd/root.go#L54

ctc_lib/config_command.go
```diff
@@ -0,0 +1,66 @@
...
+	Args:  cobra.ExactArgs(2),
+	RunE: func(command *cobra.Command, args []string) error {
+		_, exists := viper.AllSettings()[args[0]]
+		if !exists {
+			Log.Panicf("Config Key %s does not exist.", args[0])
+		}
+		// We do not want to add the default config keys in the config file.
+		// Hence we create a new instance of viper and read the config file again.
+		v := viper.New()
+		v.SetConfigFile(ConfigFile)
```
aaron-prindle: nit: might be useful to allow for config validation or potentially a callback to be run when config is set in a tool

ctc_lib/config_command_test.go
aaron-prindle: ah, makes sense.  thanks.

ctc_lib/config_command.go
tejal29: Thanks @aaron-prindle . I played with  viper.OnConfigChange() method to see if i can use it as is. 
This notification is sent after the config file written. For validation, probably we need something before the Config File is written. 
I will follow up this in a new PR as it will result in bigger diff.

PR #568, opened 2018-03-28
Change default machine type from 32CPU to 8CPU.

tomasz-safuryn:
Usually there is no need to use 32 CPUs and such a VM is very expensive.


tejal29:
LGTM.

PR #571, opened 2018-03-30
Add update check logic

tejal29:
Add Command for Update Check and check for periodic updates daily if "updatecheck" Config is set to true.

You can also override the frequency for update check by overriding the Config variable "update_check_interval_in_secs" 

Please see the Update ReadMe(https://github.com/tejal29/runtimes-common/commit/65178e02255ca96db6b18b38f88da8a14b4c45ca#diff-fb36b3e2e33055d6ec30f0bb0bdd99db)  for more details

Gist of the Code in ctc_lib/notify is copied over from [minikube/notify](https://github.com/kubernetes/minikube/tree/0fa64b3bd8b2602d41f60f2dc81dee1c451f3fd7/pkg/minikube/notify)


ctc_lib/notify/notify.go
```diff
@@ -0,0 +1,96 @@
...
+	Name      string
+	Checksums map[string]string
+}
+
+type Releases []Release
+
+func getJson(url string, target *Releases) error {
+	r, err := http.Get(url)
+	if err != nil {
+		return errors.Wrap(err, "Error getting minikube version url via http")
```
aaron-prindle: remove minikube

tejal29: done Thanks :)

dlorenc:
> LGTM. Might also make sense to add a snippet of the expected json for the update url or put the format somewhere in the docs in explicitly in addition to the minikube example link.

Yeah we put a json-schema somewhere for minikube to help with validation.

PR #576, opened 2018-04-05
Add FuncMap to Templates and validate them

tejal29:
Also change all cobra.Run to cobra.RunE so any errors when parsing and executing templates are propagated to ctc_lib.Execute() entry point.

### DIFF
--- a/ctc_lib/template_test.go
+++ b/ctc_lib/template_test.go
@@ -80,3 +80,24 @@ func TestContainerToolTestOutputWithFuncTemplate(t *testing.T) {
 		t.Errorf("Expected to contain: \n HELLO WORLD! \nGot:\n %v\n", OutputBuffer.String())
 	}
 }
+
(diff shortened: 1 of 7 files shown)


ctc_lib/template_test.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+	SetExitOnError(false)
+	err := ExecuteE(&testCommand)
+	expectedError := ("can't evaluate field greet in type ctc_lib.TestInterface")
+	if !strings.Contains(err.Error(), expectedError) {
+		t.Errorf("Expected Error to contain: \n %q \nGot:\n %q\n", expectedError, err.Error())
+	}
+}
+
+func TestContainerToolTestOutputWithFuncTemplate(t *testing.T) {
```
dlorenc: Could you add a test with a nil funcmap?

tejal29: done

PR #580, opened 2018-04-13
Add configuration from versions.yaml

jprzychodzen:


versioning/versions/versions.go
```diff
@@ -52,6 +66,50 @@ func LoadVersions(path string) Spec {
...
 }
 
+func LoadConfig(path, config string) Config {
+	var whole map[string]interface{}
+	err := yaml.Unmarshal(ReadFile(path), &whole)
+	if err != nil {
+		log.Fatalf("error: %v", err)
+	}
+
+	if c, k := whole[config]; k {
```
huyhg: replace `k` with the idiomatic `ok`

versioning/versions/versions.go
```diff
@@ -35,14 +37,26 @@ type Spec struct {
 	Versions []Version
 }
 
-func LoadVersions(path string) Spec {
+type Config map[string]string
```
huyhg: Add doc for this type

jprzychodzen: Done

versioning/versions/versions.go
jprzychodzen: Done

PR #582, opened 2018-04-13
fix cache/upload flags to work as intended

aaron-prindle:


aaron-prindle:
that was a typo.  this issue here is that currently there is no way for consumers to blowup the cache w/ FTL (--no-cache).  This is because this `cache` value:
https://github.com/GoogleCloudPlatform/runtimes-common/blob/master/ftl/python/builder.py#L57

passed to the builders also controls uploading which should only be controlled by the --no-upload flag.  As it stands, users cannot do a build not using the cache and generate cache artifacts do this condition:
https://github.com/GoogleCloudPlatform/runtimes-common/blob/master/ftl/python/layer_builder.py#L331

PR #593, opened 2018-04-17
Add json flag for ctc_lib.

tejal29:


ctc_lib/container_tool_list_command.go
```diff
@@ -76,32 +88,58 @@ Either implement Command.Run implementation or RunO implemetation`)
...
 	}
-	if err != nil {
-		return err
+	totalDisplayString, commandError, totalError = ctc.runTotalIfDefined()
+	LogIfErr(commandError, Log)
+	ctc.printResult(totalDisplayString)
+	LogIfErr(totalError, Log)
+
+	if commandError != nil && totalError != nil {
+		return errors.New("One or more errors")
```
aaron-prindle: Are errors handled the same way if json output is used?  It might be worth adding some test cases for failing commands as well in another PR.

tejal29: will add tests in this PR :) Thanks @aaron-prindle.

Issue #606, opened 2018-04-25
Parallelize build/upload for package layers


aaron-prindle:
Currently the individual package layers are uploaded sequentially, parallelizing this would speedup builds

aaron-prindle:
ref #605

Issue #605, opened 2018-04-24
FTL times out when building on Cloud Builder


myelin:
The following requirements.txt took three tries to build within the default Cloud Builder 10 minute timeout (getting a little further each time).

gunicorn
uwsgi
django
flask
google-cloud
requests
cryptography
pillow

Running this locally, it takes 17 seconds with a hot pip cache, or 61 seconds after clearing the cache with rm -rf ~/.cache/pip

Test command: time bash -c 'rm -rf env; python3 -m virtualenv -p python3 env; source env/bin/activate; pip install -r requirements.txt'

aaron-prindle:
Can you describe the environment/builder-image you are using.  In attempting the same requirements.txt I am seeing:
```
Step #2 - "build-image-gcr.io/ftl-node-test/repro_test-image:latest": INFO     Pushing image to Docker registry took 4 seconds
Step #2 - "build-image-gcr.io/ftl-node-test/repro_test-image:latest": INFO     Uploading final image took 4 seconds
Step #2 - "build-image-gcr.io/ftl-node-test/repro_test-image:latest": INFO     build process for FTL image took 392 seconds
Step #2 - "build-image-gcr.io/ftl-node-test/repro_test-image:latest": INFO     full build took 392 seconds
```

aaron-prindle:
running with w/python3.6

myelin:
Step #1 - "ftl": INFO pip_download_wheels took 281 seconds
Step #1 - "ftl": INFO build process for FTL image took 517 seconds
Step #1 - "ftl": INFO full build took 517 seconds

I'll email you the rest of the details.

aaron-prindle:
@myelin thanks for reporting this, the build times definitely rose per package when we updated to uploading each packages for phase 1.5 python.  #607 should fix this by uploading the layers in parallel. I am now seeing times:
Test requirements.txt
https://gist.github.com/aaron-prindle/758f9ef4141669e19907b17da40f657d

Timing:
https://gist.github.com/aaron-prindle/71cb60b604cd8c154205825f5b0094a0

import pieces:
pip_download_wheels took 58 seconds
uploading_all_package_layers took 13 seconds
uploading_requirements.txt_pkg_lyr took 12 seconds
full build took 122 seconds

I've made an issue with here how we might be able to drop the pip_download_wheels time:
#610

aaron-prindle:
> can the download wheels be done in parallel?
this is the overarching problem of:
[...]

PR #607, opened 2018-04-25
parallelize python phase 2

aaron-prindle:


aaron-prindle:
fixes #605

aaron-prindle:
fixes #606

aaron-prindle:
kokoro doesn't seem to be triggering for builds.  it seems that this has been the case for ~ 3 days

PR #616, opened 2018-04-26
change tar to not output root message to stderr

aaron-prindle:


aaron-prindle:
`-C / <path>` taken from https://unix.stackexchange.com/questions/59243/tar-removing-leading-from-member-names

PR #629, opened 2018-05-09
silence warning message from composer regarding running as root

aaron-prindle:


aaron-prindle:
unfortunately seems that composer writes diagnostic messages to stderr that can't specifically be turned off:
https://github.com/composer/composer/blob/3e1b0c88d7397cdd32c4ac151a8175c188ce9318/src/Composer/Command/RequireCommand.php#L144

Issue #631, opened 2018-05-09
Execute the user specified entrypoint using "sh -c"


rahulrv1980:
The user may specify an entrypoint like 'gunicorn main:app --port $PORT" and the env variable $PORT will not be expanded although the env variable is present in the docker image. it would be good to execute this via a shell to allow for shell expansion.

rahulrv1980:
ideally, this is controllable via a flag.

aaron-prindle:
fixed w/ https://github.com/GoogleCloudPlatform/runtimes-common/pull/632, closing

PR #632, opened 2018-05-10
add sh -c to all entrypoints created for env var support

aaron-prindle:


aaron-prindle:
fixes #631

ftl/common/args.py
```diff
@@ -102,6 +102,12 @@ def base_parser():
         action='store',
         default=constants.DEFAULT_ENTRYPOINT,
         help='The entrypoint for the dockerimage')
+    parser.add_argument(
```
dlorenc: Why not just do this all the time instead of as a flag?

aaron-prindle: a flag was requested by the python team, I originally did add it all the time but then it was mentioned that they might want to do it as a flag option.  They wanted this so that they could use env variables in their args

PR #661, opened 2018-06-01
make php phase 1 the default build mode for ftl php

aaron-prindle:


aaron-prindle:
I have a PR here that installs 1 pkg per layer for phase 2:
https://github.com/GoogleCloudPlatform/runtimes-common/pull/664

There is still the issue that the autoload.php files are not the same as when installed via composer install.  I don't think there is a simple way to get them to be identical.  Waiting for some response on how important the autoload files are.

aaron-prindle:
i believe it is infeasible for now as there is no way to generate the proper autoloads.  I am changing this PR to make phase_1 the default

dlorenc:
Should you just delete phase 2?

aaron-prindle:
makes sense, added a commit that deletes phase 2

Issue #622, opened 2018-05-01
Add a command to generate metadata files.


tejal29:
This command will generate 3 of the 4 metadata files and upload it to GCS at predefined location as mentioned in the TUF specification.
The 3 metadata files initially supported are
1. Root
2. Target
3. Snapshot.

PR #663, opened 2018-06-01
Minimal diff2: Add metadata structs for generating the root.json, target.json and snapshot.json

tejal29:
This is diff (2/3) for #622

In this diff,

1. Moved all constants to package constants
2. Add metadata structs for root.json, target.json and snapshot.json.
3. Tests for same.


tuf/metadata/v1/tuf_test.go
```diff
@@ -0,0 +1,274 @@
...
+	"reflect"
+	"testing"
+
+	"github.com/GoogleCloudPlatform/runtimes-common/tuf/metadata"
+	"github.com/GoogleCloudPlatform/runtimes-common/tuf/testutil"
+	"github.com/GoogleCloudPlatform/runtimes-common/tuf/types"
+)
+
+func TestRootMetadata(t *testing.T) {
+	tuf := TUF{testutil.TestTUFConfig,
```
dlorenc: nit: use the key: value initializer form.

tuf/metadata/v1/tuf.go
```diff
@@ -0,0 +1,187 @@
...
+
+type TUF struct {
+	TufConfig       config.TUFConfig
+	RootSecrets     []types.Scheme
+	OldRootSecrets  []types.Scheme
+	TargetSecrets   []types.Scheme
+	SnapshotSecrets []types.Scheme
+}
+
+func (tuf *TUF) PopulateRootMetadata() (RootMetadata, error) {
```
dlorenc: It's hard to see in this diff how these populate methods will be used. Depending on the usage, it might make more sense to do this automatically during TUF initialization. WDYT?

tejal29: Yes. Sorry, i shd have this clear in the PR description.
These will be used when someone uploads a new key using upload_secrets command.
The TODO here will be replaced by these functions. Based on which new keys are uploaded, we will re-generate all or few metadata files and upload them back to gcs.
https://github.com/GoogleCloudPlatform/runtimes-common/blob/master/tuf/deployer/deployer.go#L83.

tuf/metadata/v1/tuf_test.go
tejal29: hmm what is that? i googled but did not find an example.

tuf/metadata/v1/tuf.go
tejal29: Forgot to add, you can't do this in init because you dont have all the secrets.

tuf/metadata/v1/tuf_test.go
tejal29: done

PR #667, opened 2018-06-05
Set 6 hour TTL for unspecified package deps

### LINKED ISSUE #657: Remove TTL for package dependencies that are not fully specified
Packages that are not fully specified should not be cached so that the build done by FTL is identical to what happens locally at any given time

priyawadhwa:
Fixes #657

### DIFF
--- a/ftl/common/builder.py
+++ b/ftl/common/builder.py
@@ -102,18 +102,18 @@ class RuntimeBase(JustApp):
         cache_repo = args.cache_repository
         if not cache_repo:
             cache_repo = self._target_image.as_repository()
-        use_minimum_ttl = ftl_util.use_minimum_ttl(descriptor_files, ctx)
+        ttl = ftl_util.get_ttl(descriptor_files, ctx)
         self._cache = cache.Registry(
             repo=cache_repo,
             namespace=self._cache_namespace,
             creds=self._target_creds,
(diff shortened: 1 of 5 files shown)


ftl/common/builder.py
```diff
@@ -102,6 +102,7 @@ def __init__(self, ctx, cache_namespace, args, descriptor_files):
         cache_repo = args.cache_repository
         if not cache_repo:
             cache_repo = self._target_image.as_repository()
+        use_minimum_ttl = ftl_util.use_minimum_ttl(descriptor_files, ctx)
```
dlorenc: Maybe rename this to something that explains the why, not the what. `is_lock_file` or something similar.

ftl/common/builder.py
```diff
@@ -111,7 +112,8 @@ def __init__(self, ctx, cache_namespace, args, descriptor_files):
             mount=[self._base_name],
             use_global=args.global_cache,
             should_cache=args.cache,
-            should_upload=args.upload)
+            should_upload=args.upload,
+            use_minimum_ttl=use_minimum_ttl)
```
dlorenc: Can this be the TTL, instead of a bool?

aaron-prindle:
travis errors are related to auth change in an image we we depend on, not this PR

PR #684, opened 2018-06-15
Add new optional flag --additional-directory to add specified directory to the final built image.

rahulrv1980:


ftl/python/builder.py
```diff
@@ -113,6 +113,13 @@ def Build(self):
             exposed_ports=self._args.exposed_ports)
         app.BuildLayer()
         lyr_imgs.append(app.GetImage())
+        additional_directory = base_builder.AppLayerBuilder(
```
aaron-prindle: Right now this flag is available for all runtimes but only used in python, to fix this it would make sense to either:
only add this to python and change the flag to work in python
see: https://github.com/GoogleCloudPlatform/runtimes-common/blob/master/ftl/common/args.py#L136
OR
add this snippet and an identical test to each runtime

ftl/python/testdata/functionality_test/start.sh
```diff
@@ -1,4 +1,8 @@
 #!/bin/bash
 set -xe
-python /srv/app.py &
-wget --tries=20 localhost:8080
+if [ "$(cat /workspace/ftl/python/testdata/additional_directory/child/b.txt)" = "barfoo" ]; then
```
aaron-prindle: you shouldn't have to mess with the functionality_test, that is supposed to only test that an app runs.  For something like testing whether a folder exists, you should add this as a part of the integration tests:
https://github.com/GoogleCloudPlatform/runtimes-common/blob/master/ftl/integration_tests/ftl_python_integration_tests_yaml.py#L9

then create a folder for the test w/ a structure_test.yaml (see: https://github.com/GoogleCloudPlatform/runtimes-common/tree/master/ftl/python/testdata/packages_test)

ftl/python/builder.py
rahulrv1980: Added to node and php as well.

aaron-prindle:
lint error:
./ftl/common/args.py:49:14: E251 unexpected spaces around keyword / parameter equals

aaron-prindle:
seeing lint error:
./ftl/common/args.py:49:14: E251 unexpected spaces around keyword / parameter equals

PR #689, opened 2018-06-20
Add new, optional flag --additional-directory to package the contents of that directory into the final built image.

rahulrv1980:


aaron-prindle:
All of the cache image tests are failing.  These tests verify the number of layers that are different and make sure they match up with what is expected.  Essentially FTL expects 1 layer to be different and 2 are different due to this empty layer being added.  Can you make it so that the additional_directory layer:
https://github.com/GoogleCloudPlatform/runtimes-common/pull/689/commits/d00cd975026304c4d130d83433cb22c9a1fac1f3#diff-3694c677f32844f69656df1e1389224cR49

is only added if the flag value is present/non-empty?  Thanks.

PR #697, opened 2018-07-07
fix app layer to not include node_modules or vendor

aaron-prindle:


ftl/node/builder_test.py
```diff
@@ -63,6 +63,7 @@ def setUp(self, mock_from):
         args.base = 'gcr.io/google-appengine/python:latest'
         args.entrypoint = None
         args.tar_base_image_path = None
+        args.tar_base_image_path = None
```
priyawadhwa: nit: repeated from line 65

ftl/php/layer_builder.py
```diff
@@ -69,12 +70,13 @@ def _build_layer(self):
...
-    def _gen_composer_install_tar(self, app_dir, destination_path):
-        vendor_dir = os.path.join(app_dir, 'vendor')
-        rm_cmd = ['rm', '-rf', vendor_dir]
-        ftl_util.run_command('rm_vendor_dir', rm_cmd)
-        os.makedirs(os.path.join(vendor_dir))
+    def _cleanup_build_layer(self):
+        if self._directory:
+            vendor_dir = os.path.join(self._directory, 'vendor')
+            rm_cmd = ['rm', '-rf', vendor_dir]
+            ftl_util.run_command('rm_vendor_dir', rm_cmd)
```
priyawadhwa: Do you need 
```python
os.makedirs(vendor_dir)
````
 here? You had it after removing the node_modules dir

aaron-prindle: no it isn't required here, it's only in the builder.py methods.  The node layer_builder.py is the same, see here:
https://github.com/GoogleCloudPlatform/runtimes-common/pull/697/files

ftl/node/builder_test.py
aaron-prindle: fixed

PR #706, opened 2018-08-15
switch version check to check current version on the fly

sharifelgamal:


appengine/reconciletags/version_check_test.py
```diff
@@ -33,21 +33,64 @@
...
+    "ubuntu": "apt-get update; apt-get -y install lsb-release; lsb_release -a",
+    "ruby": "ruby -v",
+    "python": "python3 --version",
+    "php": "php -v",
+    "nodejs": "node --version",
+    "java": "java -version",
+    "go1-builder": "echo $GO_VERSION"
+}
+
+runtime_to_filter = {
```
dlorenc: For the parsing that happens outside of the docker command, could you do this in python instead? Instead of filter bash commands, this could be a map of functions?

appengine/reconciletags/version_check_test.py
```diff
@@ -13,17 +13,17 @@
 # This is the only way to import LooseVersion that will actually work
 from distutils.version import LooseVersion
 
-runtime_to_version_check = {
+runtime_to_latest_version = {
     "aspnetcore": ("git ls-remote --tags https://github.com/dotnet/core"
-                   "| egrep -o \"{}\\.[0-9]+$\" | cut -c 1-"),
+                   "| egrep -o \"{}\\.[0-9]+$\""),
```
dlorenc: Why do we still have the egreps and awks in these? I was hoping to remove them and just have a single command, then parse the rest in python.

sharifelgamal:
This should actually be ready now.

appengine/reconciletags/version_check_test.py
```diff
@@ -7,46 +7,129 @@
...
+        "nodejs": filter_node,
+        "python": filter_python,
+        "java": filter_java,
+        "ruby": filter_ruby,
+        "aspnetcore": filter_aspnetcore,
+        "go1-builder": filter_go
+    }
+
+    def _get_latest_version(self, runtime, version, image):
+        if runtime == 'java':
```
dlorenc: Why does java need a special case here? Can you add a comment?

sharifelgamal: It's the only language that needs image and version passed into the command, but turns out we can vacuously pass in those variables and just not use them for everyone else. The special case is gone.

dlorenc: Nice, much cleaner now.

PR #724, opened 2018-10-16
force server side update of composer lock for php

aaron-prindle:


voytech-net:
@aaron-prindle Hello Aaron, I have a question about this change. Our automatic builds to Google App engine started to fail because of this change. By running composer update --lock instead of composer install we run out of memory and nothing gets build. Could you please advise how we can still build our PHP Symfony app in App Engine? Thank you very much!

PR #741, opened 2018-11-30
Adds a flag 'skip_package_lock_generation' to FTL

belchatek:
Adds a flag 'skip_package_lock_generation' to control if FTL should generate package-lock file at the beginning of the run. Skiping this can improve build speed but can cause older version of dependencies to be used.

### DIFF
--- a/ftl/BUILD
+++ b/ftl/BUILD
@@ -68,6 +68,7 @@ py_test(
     srcs = ["common/util_test.py"],
     deps = [
         ":ftl_lib",
+        "@mock",
     ],
 )
 
--- a/ftl/common/args.py
+++ b/ftl/common/args.py
@@ -99,6 +99,12 @@ def base_parser():
         action='store_true',
         help='Upload to cache during build (default).')
     parser.add_argument(
+        '--skip_package_lock_generation',
+        dest='skip_package_lock_gen',
+        default=False,
+        action='store_true',
+        help='Do not create package lock file before checking cache.')
+    parser.add_argument(
         '--output-path',
         dest='output_path',
         action='store',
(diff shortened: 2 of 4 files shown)


aaron-prindle:
Nice, this change definitely makes sense.

it seems that this is failing a lint check that we have:
./ftl/node/layer_builder.py:54:80: E501 line too long (89 > 79 characters)
If you can fix that error and re submit the PR tests should run.

Also can you add an integration test for the flag that you are adding?
You will need to add the test to:
https://github.com/GoogleCloudPlatform/runtimes-common/blob/master/ftl/integration_tests/ftl_node_integration_tests_yaml.py#L14

The flag to the test similar to as done here:
https://github.com/GoogleCloudPlatform/runtimes-common/blob/master/ftl/integration_tests/ftl_node_integration_tests_yaml.py#L43

And add a directory for the app to be tested, similar to here:
https://github.com/GoogleCloudPlatform/runtimes-common/tree/master/ftl/node/testdata/packages_test

And you should create a structure_test.yaml that verifies that a package.lock file is not created:
https://github.com/GoogleCloudPlatform/runtimes-common/blob/master/ftl/node/testdata/packages_test/structure_test.yaml

belchatek:
There is a problem with adding an integration test for this flag. Even if I set it package-lock.json will be created in build phase.

belchatek:
I dont understand what is broken in kokoro run

aaron-prindle:
The kokoro issue here is related to our kokoro pipeline recently upgrading the version of bazel that it uses and the files in runtimes-common not being compatible with this new version.  We are working on a fix currently, I will update this PR as kokoro is fixed.

aaron-prindle:
@belchatek friendly ping.  The CI issues you were experiencing are resolved now, resubmit and kokoro should have sane logs for the FTL tests

PR #768, opened 2019-02-18
Fix instruction on how to install cloudbuild, dockerfiles and docgen

jprzychodzen:


docgen/README.md
```diff
@@ -4,7 +4,21 @@
 
 # How to install
 
-This uses [`bazel`](https://bazel.build) as the build tool.
+- Install packages for Debian:
```
huyhg: bazel doesn't have sufficient instructions? These instructions are also specific to debian. I would just point to bazel's official docs.

jprzychodzen: I hoped that having a copy-paste version here can be useful - we are using a specific version and instruction how to install this specific version will not change.

However, if you prefer, I will remove most of this change and leave only version-related information.

huyhg: Why the specific version? Our tools are broken with the latest Bazel build? In that case, we probably should fix the BUILD files instead.

jprzychodzen: Building is broken with latest version, as Bazel is not backward compatible.

The problem is that even if we would fix it for current version (0.23.0), and some breaking change would happen again in Bazel, then we would need to follow those changes and again fix tooling. There would be a time window between breaking and fixing, when end-user cannot use this tool, because we still didn't update BUILD files to latest version.

Fixing build tool right now would requires us to migrate from this repository, as there is specified version in build pipeline. It's shared with other projects in this repository, and this whole migration right now is a very big effort. I'm more inclined to add specific Bazel version to provide working instruction, than to follow backward-incompatible build system in new repository.

PR #783, opened 2019-03-25
Make openjdk and jetty show up as the java runtime to clarify output

sharifelgamal:
The bug I'm fixing was introduced in #773. 

The file we're parsing for both openjdk and jetty is called java.json, but we need to check them entirely separately, so I reuse the runtime variable to filter on the proper runtime. 

Once we're done checking and aggregate our finding, I want them both back under the "java" heading, so this is just me changing it back.

### DIFF
--- a/appengine/reconciletags/version_check_test.py
+++ b/appengine/reconciletags/version_check_test.py
@@ -208,6 +208,8 @@ class VersionCheckTest(unittest.TestCase):
                                 }
                                 old_images.append(entry)
                 if old_images:
+                    if runtime == 'jetty' or runtime == 'openjdk':
+                        runtime = 'java'
                     images_map[runtime] = old_images
 
         if images_map:


appengine/reconciletags/version_check_test.py
```diff
@@ -208,6 +208,8 @@ def test_latest_version(self):
                                 }
                                 old_images.append(entry)
                 if old_images:
+                    if runtime == 'jetty' or runtime == 'openjdk':
```
tstromberg: I can't make the connection between this code change and the current PR description. The code seems like it should work, but I can't review the PR without some more context. 

These are the kind of context questions I look to be answered in a PR description:

- Is there a bug that this CL addresses?
- What is the problem that's being solved?
- Why this is the best approach?

sharifelgamal: The bug I'm fixing was introduced in https://github.com/GoogleCloudPlatform/runtimes-common/pull/773. The file we're parsing for both openjdk and jetty is called java.json but we need to check them entirely separately, so I reuse the runtime variable to filter on the proper runtime. Once we're done checking and aggregate our finding, I want them both back under the "java" heading, so this is just me changing it back.

PR #793, opened 2019-05-09
Added --add-directories flag

ovk6:
Added possibility to create directories for new versions with --add-directories flag

### DIFF
--- a/versioning/scripts/dockerfiles/main.go
+++ b/versioning/scripts/dockerfiles/main.go
@@ -9,6 +9,7 @@ import (
 	"io/ioutil"
 	"log"
 	"os"
+	"path"
 	"path/filepath"
 	"reflect"
 	"strings"
@@ -101,13 +102,16 @@ func findFilesToCopy(templateDir string, callback func(path string, fileInfo os.
 }
 
 func copyFiles(version versions.Version, templateDir string) {
-	findFilesToCopy(templateDir, func(path string, fileInfo os.FileInfo) {
-		data, err := ioutil.ReadFile(filepath.Join(templateDir, path))
+	findFilesToCopy(templateDir, func(filePath string, fileInfo os.FileInfo) {
+		data, err := ioutil.ReadFile(filepath.Join(templateDir, filePath))
 		check(err)
 
-		target := filepath.Join(version.Dir, path)
+		target := filepath.Join(version.Dir, filePath)
 		// Delete first to make sure file is created with the right mode.
 		deleteIfFileExists(target)
+		// Create nested directory structure if needed.
+		dirPath := path.Dir(target)
+		os.MkdirAll(dirPath, os.ModePerm)
 		err = ioutil.WriteFile(target, data, fileInfo.Mode())
 		check(err)
 	})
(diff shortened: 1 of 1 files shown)


ovk6:
@huyhg please review

versioning/scripts/dockerfiles/main.go
```diff
@@ -101,13 +102,16 @@ func findFilesToCopy(templateDir string, callback func(path string, fileInfo os.
...
+		data, err := ioutil.ReadFile(filepath.Join(templateDir, filePath))
 		check(err)
 
-		target := filepath.Join(version.Dir, path)
+		target := filepath.Join(version.Dir, filePath)
 		// Delete first to make sure file is created with the right mode.
 		deleteIfFileExists(target)
+		// Create nested directory structure if needed.
+		dirPath := path.Dir(target)
+		os.MkdirAll(dirPath, os.ModePerm)
```
wgrzelak: Why do we need to create directory twice?

Please see https://github.com/ovk6/runtimes-common/blob/6a52810179b21698276ab1cb2c12c44f680f25c0/versioning/scripts/dockerfiles/main.go#L259.

versioning/scripts/dockerfiles/main.go
```diff
@@ -227,6 +231,7 @@ func check(e error) {
 func main() {
 	defaultTemplateDirPtr := flag.String("template_dir", "templates", "Path to directory containing Dockerfile.template and any other files to copy over")
 	verifyPtr := flag.Bool("verify_only", false, "Verify dockerfiles")
+	addDir := flag.Bool("add_directories", false, "Add new directories")
```
wgrzelak: This should be `addDirPtr` instead.

wgrzelak: I think `create_directories` is more verbose.

versioning/scripts/dockerfiles/main.go
ovk6: Yes, because in current configuration it was not providing nested directory structure.

Eg. in case there is nested directory:
```
templates/Dockerfile.template
templates/configs/xxx.conf
```

it was failing if directory `configs` was not created in target directory.

```
$ dockerfiles
panic: open 7/debian9/7.1/configs/xxx.conf: no such file or directory
```

wgrzelak: It make sense, but this the directory will be created always. I think we wanted to use the `--add_directories` flag or rather `--create_directories`?

versioning/scripts/dockerfiles/main.go
```diff
@@ -250,6 +255,9 @@ func main() {
 			failureCount += verifyDockerfiles(version, filepath.Join(*defaultTemplateDirPtr, version.TemplateSubDir), *tmpl)
 			failureCount += verifyCopiedFiles(version, filepath.Join(*defaultTemplateDirPtr, version.TemplateSubDir))
 		} else {
+			if *addDir {
```
wgrzelak: Comparing on the comments from `copyFiles` func, I think we should move this to the `writeDockerfile` to be consistent on what the `copyFiles` function does.

versioning/scripts/dockerfiles/main.go
wgrzelak: @ovk6 please rename the argument.

versioning/scripts/dockerfiles/main.go
ovk6: Yes, but I think it is a bug in original software. This directory structure should be copied even if we don't want to create new directories for new versions.

wgrzelak: Please correct me if I'm wrong. Will this create the directory contains `version.Dir` value even if `--add_directories` flag is `false`?

PR #795, opened 2019-06-18
Renamed flag, fixed default behavior

ovk6:
Renamed add_directory to create_directory
Fixed to only create dirs when flag is passed for backward compatibility

### DIFF
--- a/versioning/scripts/dockerfiles/main.go
+++ b/versioning/scripts/dockerfiles/main.go
@@ -80,10 +80,14 @@ func renderDockerfile(version versions.Version, tmpl template.Template) []byte {
 	return result.Bytes()
 }
 
-func writeDockerfile(version versions.Version, data []byte) {
+func writeDockerfile(version versions.Version, data []byte, createDirPtr *bool) {
 	path := filepath.Join(version.Dir, "Dockerfile")
 	// Delete first to make sure file is created with the right mode.
 	deleteIfFileExists(path)
+	// Create nested directory structure if needed.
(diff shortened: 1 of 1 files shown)


versioning/scripts/dockerfiles/main.go
```diff
@@ -231,7 +236,7 @@ func check(e error) {
 func main() {
 	defaultTemplateDirPtr := flag.String("template_dir", "templates", "Path to directory containing Dockerfile.template and any other files to copy over")
 	verifyPtr := flag.Bool("verify_only", false, "Verify dockerfiles")
-	addDir := flag.Bool("add_directories", false, "Add new directories")
+	createDirPtr := flag.Bool("create_directories", false, "Create new directories")
```
wgrzelak: Can you simply move the flag to be a global one? So, will be not necessary to pass the flag via argument.

Example: https://github.com/GoogleCloudPlatform/runtimes-common/blob/62fe772429d1633e594e66e2549c23051274e767/docgen/scripts/docgen/docgen.go#L12

ovk6: It is more readable to have options parsing all in one place, not scattered some globally and some here.
I think there is nothing wrong in passing it as argument.

Btw, in example you send there is only one place where variable is used - main function, so there is no justification to make it global.

What do you think?

wgrzelak: Then, please pass the value as value, not a pointer.

PR #801, opened 2020-07-27
Added image builder support

aav66:


versioning/scripts/cloudbuild/main.go
```diff
@@ -214,10 +240,20 @@ func newCloudBuildTemplateData(
...
 		}
-		data.AllImages = append(data.AllImages, images...)
+		// Ignore bulder images from images list
+		if !v.Builder {
+			data.AllImages = append(data.AllImages, images...)
+		}
 		versionSTests, versionFTests := filterTests(structureTests, functionalTests, v)
-		data.ImageBuilds = append(
-			data.ImageBuilds, imageBuildTemplateData{v.Dir, images[0], images[1:], versionSTests, versionFTests})
+		// Enforce to use ImageNameFromBuilder as referecne to create tags
```
armandomiani, line 248: reference

versioning/scripts/cloudbuild/main.go
```diff
@@ -214,10 +240,20 @@ func newCloudBuildTemplateData(
 				break
 			}
 		}
-		data.AllImages = append(data.AllImages, images...)
+		// Ignore bulder images from images list
```
armandomiani, line 243: builder

versioning/scripts/dockerfiles/main.go
```diff
@@ -244,6 +244,10 @@ func main() {
 	spec = versions.LoadVersions("versions.yaml")
 
 	for _, version := range spec.Versions {
+		// Ignore version without Dir for possibility to use builder images
+		if version.Dir != "" {
+			break
```
armandomiani, line 249: break or continue?

aav66: For sure continue. Fixed.

versioning/scripts/cloudbuild/main.go
```diff
@@ -214,10 +240,20 @@ func newCloudBuildTemplateData(
...
 		}
-		data.AllImages = append(data.AllImages, images...)
+		// Ignore builder images from images list
+		if !v.Builder {
+			data.AllImages = append(data.AllImages, images...)
+		}
 		versionSTests, versionFTests := filterTests(structureTests, functionalTests, v)
-		data.ImageBuilds = append(
-			data.ImageBuilds, imageBuildTemplateData{v.Dir, images[0], images[1:], versionSTests, versionFTests})
+		// Enforce to use ImageNameFromBuilder as reference to create tags
```
ovk6, line 248: What is the purpose of ImageNameFromBuilder variable?

versioning/scripts/cloudbuild/main.go
```diff
@@ -64,6 +65,27 @@ const cloudBuildTemplateString = `steps:
     waitFor: ['-']
     id: 'image-{{ .Tag }}'
 {{- end }}
+{{- else }}
+{{- if .BuildOn }}
```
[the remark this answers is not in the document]
aav66: First two is accepted.

versioning/versions/versions.go
```diff
@@ -25,14 +25,18 @@ type Package struct {
...
-	ExcludeTests   []string `yaml:"excludeTests"`
+	Dir                  string
+	TemplateSubDir       string `yaml:"templateSubDir"`
+	Repo                 string
+	Tags                 []string
+	From                 string
+	TemplateArgs         map[string]string `yaml:"templateArgs"`
+	Packages             map[string]Package
+	ExcludeTests         []string `yaml:"excludeTests"`
+	Builder              bool
```
huyhg, line 36: Instead of using a boolean here, just check the value of the `BuilderImage` instead: its being non-empty should be equivalent to this boolean set to `true`.

versioning/scripts/cloudbuild/main.go
aav66: It is used. It is need because image builder can create image with different name. this image name used as a reference to create new tags.

versioning/versions/versions.go
aav66: Fixed

versioning/scripts/cloudbuild/main.go
huyhg: Consumers of `imageBuildTemplateData` don't appear to use that field though, so perhaps it should be removed there and only kept in the `Version` struct.

versioning/versions/versions.go
```diff
@@ -25,14 +25,18 @@ type Package struct {
...
-	ExcludeTests   []string `yaml:"excludeTests"`
+	Dir                  string
+	TemplateSubDir       string `yaml:"templateSubDir"`
+	Repo                 string
+	Tags                 []string
+	From                 string
+	TemplateArgs         map[string]string `yaml:"templateArgs"`
+	Packages             map[string]Package
+	ExcludeTests         []string `yaml:"excludeTests"`
+	Builder              string
```
huyhg, line 36: Optional: I think it's better to move these builder related fields into a nested struct instead.

versioning/scripts/cloudbuild/main.go
aav66: They use it. This field used as original tag which needed to create other tags for image.

versioning/versions/versions.go
aav66: I think it will be better to keep it as is. For each version it can be one or two additional fields used. It will be look not good for create nested only one filed or two.

PR #853, opened 2026-03-21
fix: prevent PyPI resolution for error_test fixture (dependency confusion)

x4v13r64:
`ftl/python/testdata/error_test/requirements.txt` contains a single entry, `floogle`, which is not registered on PyPI. The file is an error-handling test fixture, so a non-existent package name is intentional — but without any index restriction, pip will query PyPI for it.

An attacker who registers `floogle` on PyPI would have it installed whenever a developer or CI pipeline runs `pip install -r` against this fixture.

This PR adds `--no-index` to the file so that pip cannot resolve the package from PyPI, preserving the test intent while eliminating the squatting risk.


x4v13r64:
@sharmaharisam can you review please?

